PR #120, opened 2014-08-12
Added TurnVec to LuaUnitScript

PicassoCT:
I hope im doing this right.

Sorry if something is not as it suppossed to be. :(


abma:
@PicassoCT:

> why cant i submit a pull request to springrts.com?

i guess because this pull request already exists?!

PicassoCT:
I think i fixed it now :(
Sorry for the going back and forth..
Never got spring to compile under winbooze.. shouldnt do this  i know, but this was after all planned only as a little patch.

Kaylewt:
Like your posts, 99% of these changes are nothing but chaotic noise. Please do not touch critical code this way.

PicassoCT:
@Kaylewt 
Hate to admit it, but you are right. This is clearly not a clean comit and the fact that several patches are tangled up in it (one of them beeing actually discovered by jk )

Should get this compiled and checked before committing (also every section of a engine is by definition critical. Thus mantis must be empty.  Also just because im a gamedev - sound engineering is not to be expected? ) .

Regarding PathNode Struc and purely chaotic noise. Yes, rearranging did not change the padding, my mistake to assume there was inter-struct padding. There is no padding between structs.

First thing:
http://goo.gl/xqya3d
Notice how the inline instruction gets ignored. This happens even if functions are used. The compiler decides to add those functions to the instruction stream instead of having them as 32 bit functionpointers in the struct.

The struct boils down to 5 members of 4 byte each - totalling at 20 bytes, which on a common 64 bit wide system means  one part of the last 64 bit word is used by the next PathNode allready.
■■ ■■ ■o  
aa aa ab bb bb cc cc cd dd dd

So what to do?
Either boil it down to 16 bytes (which would be ideal for 32-bit systems and 64 bit systems) or add a unsued 32 bit int, to bring it to at least a bus-sized load on 64bit.

And yes, im doing a rewrite of the whole storage for pathnodes. Deal with it. Will be interesting to see who profiles better.

PR #128, opened 2014-09-04
Propose Spring to render same world twice to assist stereoscopic widget (with a working code)

xponen:
There is a stereoscopic widget that emulate 3D vision by moving camera left & right ( https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaUI/Widgets/gfx_stereo3d.lua ) . But the method used by this widget cannot work properly for moving units. If Spring could render same world twice then that method could work!

I have tried it, here is the result with rendering twice:
(try to focus on the circular white cursor using cross-eye technique)
stereoscopic- with doubledraw

The following is the same scene but without rendering twice:
stereoscopic- no doubledraw

The boat in the centre is moving and it doesn't 3D correctly in second image.

In the proposed code-change I added the following Interface:
/doubledraw 1|0 : enable/disable double drawing
Spring.IsDoubleDrawing(): to check if double drawing is active
Spring.GetDoubleDrawIndex(): output 1 or 2 to assist widget to select which draw-frame to displayed as left-eye or right-eye

### DIFF
--- a/rts/Game/Game.cpp
+++ b/rts/Game/Game.cpp
@@ -214,6 +214,8 @@ CR_REG_METADATA(CGame, (
 	CR_MEMBER(playing),
 	CR_IGNORED(msgProcTimeLeft),
 	CR_IGNORED(consumeSpeedMult),
+	CR_IGNORED(doubleDrawIndex),
+	CR_IGNORED(doDoubleDraw),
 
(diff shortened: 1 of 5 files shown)


silentwings:
2014 was a fine vintage of PRs but its time has passed

PR #132, opened 2014-10-03
Remove random undocumented beamlaser range modifer

sprunk:
Remove beamlaser 30% range boost on mobile units.
~~Remove beamlaser -5% range penalty when FPSed.~~

### DIFF
--- a/rts/Sim/Weapons/BeamLaser.cpp
+++ b/rts/Sim/Weapons/BeamLaser.cpp
@@ -270,11 +270,16 @@ void CBeamLaser::FireImpl(bool scriptCall)
 void CBeamLaser::FireInternal(float3 curDir)
 {
 	float actualRange = range;
+	float rangeMod = 1.0f;
+
+	if (owner->UnderFirstPersonControl()) {
+		rangeMod = 0.95f;
+	}
 
 	bool tryAgain = true;
 	bool doDamage = true;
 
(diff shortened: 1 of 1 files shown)


GoogleFrog:
The -5% is because you can FPS units to fire slightly beyond their range or graze the edge of hitvolumes of units which are technically not in range. This is most important when a turret is involved.

I agree with removing the +30% for mobile units. You can overshoot with inaccurate weapons force fired at the ground or poorly placed enemy units.

sprunk:
Grazing out-of-range hitvolumes will always be a problem as long as targetBorder < 1.0 and disappears with targetBorder = 1.0; for any range reduction value and targetBorder pair you can give a range and hitvolume radius pair that allows grazing, which is any hit radius larger than ((1-reduction) / (1-targetBorder))*range. This is related to the weapondef's targetBorder setting and unrelated to FPS (would be possible to do with normal orders too if you could give an attack order in the air - it can possibly even be reproduced now for certain setups).

GoogleFrog how recently have you checked it for turrets specifically? When BA disallowed FPSing turrets the 130% was for every beamlaser (ie. turrets too) and there was no 95% override for FPS; the +30% was the cause for range hax.

sprunk:
Updated to only remove the +30% for mobiles.

sprunk:
Looks like not happening.

PR #133, opened 2014-10-04
add dynamic damage tags to WeaponDefs

sprunk:
dynDamageExp
dynDamageMin
dynDamageRange
dynDamageInverted


gajop:
This PR needs to be cleaned up: too many commits for so few changes. Also, I'm not familiar with tags, but how are these used exactly?

sprunk:
These are existing weapon defs tags, the patch is to expose them for reading by Lua nuggets.
Regarding cleanup: replacing this PR with #181.

PR #148, opened 2015-01-02
Fix shadow fbo typo

rlcevg:


abma:
my opengl knowledge is limited.... can you explain in short what was wrong and why this is correct?

i'll give this change a try, thanks!

rlcevg:
related thread: http://springrts.com/phpbb/viewtopic.php?f=11&p=564257#p564257
Consider example in https://www.opengl.org/wiki/Framebuffer_Object_Examples [snippet "Quick example, render_to_texture (2D Depth texture ONLY)"]
Appears some permisive glsl compilers are fine with only glDrawBuffer(GL_NONE) (notice how its duplicated twice without fix? = typo), it works fine with linux's glsl 4.4 for example. But OS X's glsl 1.2 was more strict and gave nex error:
[f=0000000] Warning: FBO-SHADOW: Incomplete read buffer!

abma:
aah, great, thanks!

PR #149, opened 2015-01-03
Improving configuration for mingw

12345swordy:
Creating improvements to the build environment for mingw64 tool chain. This may result a more easier setup for new window developers.


abma:
if we have at some point 64bit mingwlibs, how can this be still used?

also the message:
"ERROR: strnlen fuction is not found"
should be changed to something useful. what can be done that strnlen is be found/what is wrong? wrong compiler? wrong headers? ... whats wrong ?

abma:
also ident seems broken, tab & spaces are used, please use tab only.

12345swordy:
This is a temp fix, until the mingwlib is updated to 64bit.
I can change the error message from "ERROR: strnlen function is not found" to "ERROR: strnlen function is not found, your compiler does not have funciton in the <string.h> file"
Is that more meaning full? I edit it to reflect this.
"also ident seems broken, tab & spaces are used, please use tab only."
huh, it looks fine when I edit that, will fix that.

abma:
> This is a temp fix, until the mingwlib is updated to 64bit.
very likely mingwlibs never will be "updated" to 64 bit. very likely mingwlibs64 will be created/added.

12345swordy:
Who's going to create mingwlib64 then?

abma:
idk as benefits are pretty low and its a lot of work. as long as spring doesn't hit the 4GB memory limit and doesn't rely on 64bit features i don't see this happen soon. having only 64 bit (official) binaries for linux/osx wouldn't be a big deal, not sure how many users on windows still have 32 bit.

abma:
wiki page should be updated. this change possible breaks other compilers. also the message it gives is quiet clear: "Spring's MinGWLibs don't include 64bit libraries (yet). Please, compile as 32bit."

PR #151, opened 2015-01-05
Update UnitScript.cpp

PicassoCT:
Because the original Error Messages were not really helpfull..


abma:
please use const std::string (instead of string)

abma:
also doesn't compile:

https://travis-ci.org/spring/spring/jobs/45959169

jk3064:
```c++
string DbgString="emit-sfx: GetEmitDirPos failed at Unit:"+unit->id +" of type "+unit->UnitDefID;
```
c++ doesn't work like that, you need printf() or IntToString()

PR #153, opened 2015-01-11
Linux Stacktrace()

rlcevg:
Stacktrace () will not terminate itself on DebugGLStacktraces=1, and should work in similar way to old CrashHandler.
https://github.com/spring/spring/blob/91c9870e5d3e34490f4849cf902e7277c0de4b64/rts/System/Platform/Linux/CrashHandler.cpp#L470

related: http://springrts.com/mantis/view.php?id=4646


abma:
thanks a lot!

fyi: https://github.com/spring/spring/commit/4abe4160dbe1ac37c545c485315bf82cb285e4c4 was needed to make it work :)

PR #155, opened 2015-01-16
Lua lock changes

ashdnazg:
I've currently put a test instead of the lock checking if it's called from the main thread and logging an error if it isn't.
I haven't seen any error yet in both windows and linux.
We may want to surround the #define s with #ifndef NDEBUG so it isn't tested on releases.

### DIFF
--- a/rts/lib/lua/include/LuaUser.cpp
+++ b/rts/lib/lua/include/LuaUser.cpp
@@ -29,10 +29,10 @@ static boost::recursive_mutex* GetLuaMutex(lua_State* L)
 void LuaCheckIfMain(lua_State* L)
 {
 
-	if (!Threading::IsMainThread()) 
-    {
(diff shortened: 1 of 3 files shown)


abma:
thanks a lot!

imo assert(Threading::IsMainThread()); would be better as it would abort/crash spring & create a stacktrace (so it never can be missed when it happens). but thats very minor.

@jk3064 
thoughts? merge?!

ashdnazg:
I thought about assert too, I'm alright with whatever you guys prefer.

Anarchid:
If not mergeable into develop it would still in my unqualified opinion be interesting to at least segregate it into a separate yet built branch, as was done with syncdebug and others in the past.

ashdnazg:
https://github.com/spring/spring/tree/LockFreeLua

Branch was set up, and kloot has implemented the changes, so this is now closed.

PR #157, opened 2015-01-23
Fix InfoTexture bugs

db81:
Clears flags cashe in GLSLProgramObject::Release() so that when you call LoadFromLua() shaders are recompiled properly.

This fixes InfoTextures not working with the error:
```
    [f=0000000] [Shader] Warning: [GLSL-PO::Link] program-object name: CInfoTextureCombiner, link-log:
    Link info
    ---------
    No shader objects attached.
     
    [f=0000000] Error: CInfoTextureCombiner-shader compilation error: Link info
    ---------
    No shader objects attached.
```


ashdnazg:
Also should fix http://springrts.com/mantis/view.php?id=4649

PR #158, opened 2015-01-24
Update LockFreeLua

ashdnazg:


ashdnazg:
it's overwriting the LockFreeLua changes, will make a new one

PR #160, opened 2015-01-24
Fix modrules experience scaling values being overridden by hardcoded values

sprunk:
Modrules already set these things and handle the default values.
Setting them in Unit.cpp again to hardcoded values overrided the modrules.


abma:
thanks! kloot cleaned it a bit up:

https://github.com/spring/spring/commit/cae510ffc8dfa8907a7abdbf5912a12637f8eae6

PR #161, opened 2015-01-24
Unitsync crash fix

ashdnazg:
(also added the new experience fix by kloot)

### DIFF
--- a/rts/Game/GameVersion.cpp
+++ b/rts/Game/GameVersion.cpp
@@ -178,6 +178,15 @@ bool IsHeadless()
 #endif
 }
 
+bool IsUnitsync()
+{
+#ifdef UNITSYNC
+	return true;
+#else
+	return false;
+#endif
+}
+
 const std::string& Get()
 {
 	static const std::string base = IsRelease()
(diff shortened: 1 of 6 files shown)


abma:
[21:23:51] <[LCC]jK> doesn't work
[21:23:59] <[LCC]jK> you forget that unitsync is a library
[21:24:06] <[LCC]jK> and that the lobby calls the functions
[21:24:21] <[LCC]jK> and who says that the lobby calls them always from the same thread?
[22:16:49] <[AG]abma> imo for unitsync enable full check

abma:
sorry, my idea is wrong: liblua is only compiled once, so preprocessor doesn't work. (thanks jk for this hint)

abma:
what about adding an "bool IsUnitSync()" to GameVersion.h/GameVersion.cpp and use that in the assert?

compiling liblua twice imo pretty bad: more codepaths, longer compile time, duplicate file names list in CMakeLists.txt, -> more difficult to test.

SetMainThread() didn't work, because it stores current threadid and this assert will fail when called from different thread still calling from a single thread only. (still unitsync is far away from beeing threadsafe)

abma:
thanks!

PR #164, opened 2015-02-03
Update Lock Free Lua mainly for whitening fix

GoogleFrog:


abma:
imo efforts should go into https://github.com/spring/spring/pull/165

please a bit patience :-)

(waiting for feedback of jk)

PR #165, opened 2015-02-03
Merge LockFreeLua to develop

ashdnazg:
Lock yo' wives and hide yo' daughters


abma:
imo give this change a try even if it possible a micro-optimization. atm it looks like benefits are greater than possible drawbacks.

PR #166, opened 2015-02-04
Fix User target being set for all weapons

ashdnazg:
haveUserTarget is already set in
https://github.com/spring/spring/blob/868b2a0bdcbae486553838509ee97ad4dc5808ac/rts/Sim/Weapons/Weapon.cpp#L622
or
https://github.com/spring/spring/blob/868b2a0bdcbae486553838509ee97ad4dc5808ac/rts/Sim/Weapons/Weapon.cpp#L654

When set also in the change location, all weapons automatically get haveUserTarget status even if they target a different target (since they can't target the user one).


jk3064:
what is when the usertarget becomes targetable, is haveUserTarget correctly updated afterwards with your change?

ashdnazg:
Ok, I've tested and the flag is correctly updated for the relevant weapons. (becomes true when target unit is in range).

PR #167, opened 2015-02-13
[AI] Add Map::GetWaterDamage

rlcevg:
Currently there is no nice way to identify if water is appropriate place for buildings, except for parsing map's .smd
http://springrts.com/mantis/view.php?id=4676

SIDENOTE: Does next code know that water.damage is already scaled by ((float)UNIT_SLOWUPDATE_RATE / GAME_SPEED)?
@see https://github.com/spring/spring/blob/36f68e956d0e6ca45b03d9dc27d717084c43503e/rts/Sim/MoveTypes/MoveDefHandler.cpp#L130-L132
I think water.damage for that part of code should be scaled back with (GAME_SPEED/UNIT_SLOWUPDATE_RATE).

### DIFF
--- a/rts/ExternalAI/SSkirmishAICallbackImpl.cpp
+++ b/rts/ExternalAI/SSkirmishAICallbackImpl.cpp
@@ -1912,6 +1912,10 @@ EXPORT(float) skirmishAiCallback_Map_getGravity(int skirmishAIId) {
 	return skirmishAIId_callback[skirmishAIId]->GetGravity();
 }
 
+EXPORT(float) skirmishAiCallback_Map_getWaterDamage(int skirmishAIId) {
+	return mapInfo->water.damage;
+}
+
 
 
 EXPORT(bool) skirmishAiCallback_Map_isPossibleToBuildAt(int skirmishAIId, int unitDefId,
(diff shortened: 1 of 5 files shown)


rts/ExternalAI/SSkirmishAICallbackImpl.cpp
```diff
@@ -1912,6 +1912,10 @@ EXPORT(float) skirmishAiCallback_Map_getGravity(int skirmishAIId) {
 	return skirmishAIId_callback[skirmishAIId]->GetGravity();
 }
 
+EXPORT(float) skirmishAiCallback_Map_getWaterDamage(int skirmishAIId) {
+	return mapInfo->water.damage * ((float)GAME_SPEED / UNIT_SLOWUPDATE_RATE);
```
abma: imo changing water.damage in SSkirmishAICallbackImpl.cpp is wrong. it should return the "plain" water.damage. shouldn't ai only care if water.damage > 0 or water.damage <= 0?!

rlcevg:
Code from https://github.com/spring/spring/blob/36f68e956d0e6ca45b03d9dc27d717084c43503e/rts/Sim/MoveTypes/MoveDefHandler.cpp#L130-L132 reveals that there is more than just >0 or <=0. It also shows the buggy range of water.damage where hovers are actually usable even if water.damage > 0.
And actually water.damage is not plain, its already scaled: https://github.com/spring/spring/blob/36f68e956d0e6ca45b03d9dc27d717084c43503e/rts/Map/MapInfo.cpp#L226
I don't like scaling in SSkirmishAICallbackImpl.cpp too but can't see the other way (except maybe duplicating struct damage member inside water_t).
Related: http://springrts.com/mantis/view.php?id=4675

Probably should close this PR (with http://springrts.com/mantis/view.php?id=4676), and first create specification for new Map interface, because now it lacks more than WaterDamage. For instance i'd like to access also mapHardness and notDeformable params. @abma, if you agree close it.

rlcevg:
On second thought i'll remove scaling from SSkirmishAICallbackImpl.cpp
Because having >0 or <=0 is better then nothing. And if water.damage param came from lua and not .smd, then parsing .smd won't work.

abma:
without scale back imo its ok. thanks!

PR #168, opened 2015-02-14
allow different tile sizes to be loaded

enetheru:
Hi,

Let me know what needs fixing and I will alter it to match.


enetheru:
here's some test files you can use to load maps with tilesizes of 32,64,128,256,512,1024
https://www.dropbox.com/sh/bd305aqv8t3k3jy/AADp6ZzLCLiYVBpZPvitoCBAa?dl=0

enetheru:
That's what i get for editing on github and not testing.

enetheru:
i'll re-work and resubmit at a later date.

PR #169, opened 2015-02-16
Slow update change

GoogleFrog:
This PR is to fix incorrect self-d timers as well as the incorrect tooltip and unitdefs interpretations across the entire economy. I would like technical feedback because I have had no negative feedback yet but it is likely that it exists.

Forum thread: https://springrts.com/phpbb/viewtopic.php?f=12&t=33041


abma:
thanks, lets see if we find bugs/issues :)

PR #172, opened 2015-03-05
Added and fixed some lua for projectiles.

GoogleFrog:


rts/Lua/LuaSyncedCtrl.cpp
```diff
@@ -2792,17 +2795,17 @@ int LuaSyncedCtrl::SetProjectileTarget(lua_State* L)
 	switch (lua_gettop(L)) {
 		case 3: {
 			const int id = luaL_checkint(L, 2);
-			const int type = luaL_checkint(L, 3);
+			const string type = luaL_checkstring(L, 3);
 
 			CWorldObject* oldTargetObject = wpro->GetTargetObject();
 			CWorldObject* newTargetObject = NULL;
 
-			switch (type) {
```
abma: whats wrong with switch/case here? imo switch/case should be prefered over if/else when possible.

rts/Lua/LuaSyncedCtrl.cpp
```diff
@@ -2792,17 +2795,17 @@ int LuaSyncedCtrl::SetProjectileTarget(lua_State* L)
 	switch (lua_gettop(L)) {
 		case 3: {
 			const int id = luaL_checkint(L, 2);
-			const int type = luaL_checkint(L, 3);
+			const string type = luaL_checkstring(L, 3);
```
abma: oops, sorry didn't see that

GoogleFrog:
Anything else?

Anarchid:
> @gajop 6 hours ago:  I'd click pull but not sure why you added allyTeam info in the projectile, you sure you didn't break something? [^](http://zero-k.info/Forum/Thread/11835#121181)

gajop:
just to add this (sorry for posting off-PR):
@GoogleFrog:
The allyTeam info is there because otherwise projectiles depend on their owner unit to check allyTeam for antinuke and shield interception. It seems that this works even for dead owners, the object is not destroyed? The owner is just used for allyTeam and wobble reduction based on XP for missiles.

I wanted to reconstruct nuke/antinuke interactions so the allyTeam of the projectiles needed to be set somehow. Setting a number in the projectile is much simpler than creating some owner unit to have an allyTeam. Also I think it makes sense for projectiles to have the allyTeam of the unit when it was fired.

PR #174, opened 2015-03-06
Make caching of load times per game per map

DeadnightWarrior:
If you play the same map on different games, the load times vary a lot, and progress bar is usualy wrong.


ashdnazg:
If I'm not mistaken, try the following:
1) Delete the cache.
2) Start a game on a map.
3) Start a different game on a map (not necessarily the same map)
4) Check infolog for trying to index nil error.

DeadnightWarrior:
Did you check that games don't have their own implementation of loadprogress.lua which is not compatible with this one? LUA stuff in games usualy overrides the one in springcontent.sdz. So if both want to access the same cache file, which had different than expected structure, it'll cause nil errors.

BA and XTA for example have LuaIntro/Addons/loadprogress.lua which overrides the one in springfiles. Even though the contents of those files is identical to the one in springfiles.sdz, which means, games don't need to have their own version of this.

Anarchid:
Some of this looks like it could use some squishing

ashdnazg:
I actually didn't check before writing the comment, but I don't feel too bad since you haven't checked either.
Anyway, since you didn't test the code I had to do this myself:

[f=0000000] [LuaIntro] Error: Failed to load: loadprogress.lua  (error = 2, LuaIntro/Addons/loadprogress.lua, [string "LuaIntro/Addons/loadprogress.lua"]:20: attempt to index field '?' (a nil value))
[f=0000000] [LuaIntro] Warning: Missing/Unloaded dependency "LoadProgress" for "Main".
[f=0000000] [LuaIntro] Warning: Missing/Unloaded dependency "LoadProgress" for "Music".
[f=0000000] [LuaIntro] Warning: Missing/Unloaded dependency "LoadProgress" for "LoadTexture".

You only initialise the game's inner table when the file doesn't exist, and not when the game wasn't played yet.

abma:
maybe see for some of the reasons why it is as it is: https://github.com/spring/spring/commit/e3e11ab94f9ca1a08fbe39945612ba55a326ccd7

and yes, this should be a single commit and 12 commits (!!!).

DeadnightWarrior:
ashdnazg is correct about his remark. I havent tested it with other games since most have custom load screen, and that would mean extracting their archive and overwriting the file in question, which I didn't bother with.

I haven't made a pull request on Git in like 2 years, so I forgot to push develop branch merge, before pushing this change.

Either way, it needs initialization of inner table in case when file structure exists, but doesn't contain entries from a new game. I'll fix it, and test this time.

Regarding jk comments on abma's implementation, my changes are nowhere near as elaborate as that, and shouldn't cause any extra maintenance.

abma:
can you please create a new pull request which doesn't contain 13 commits to change 6 lines?

PR #177, opened 2015-03-21
CMakeLists: add compile option that allow me to compile in MS Windows

xponen:
I added new options & flags that allow me (and hopefully others) to successfully compile and run Spring in Windows.

I'm using
1) Window7 64-bit
2) TDM_GCC_64-bit version 4.9.2 
3) CMAKE version 3.2.1
4) Spring at https://github.com/spring/spring/commit/8ed70de60b0de05067fd1278ec58b127c4539a06
5) MinGWLib at https://github.com/spring/mingwlibs/commit/d7cea7396931633fde05035c3c9d0f395817def3
6) new build option: "FASTBUILD"

I also added CMAKE echo to suggest users to use TDM_GCC_64-bit if they tried to compile Spring using 32-bit MinGW (TDM_GCC_32-bit). Which caused "strnlen not defined in RapidHandler.cpp" error <--although one could just fix this with local function like this:

size_t strnlen(const char* s, size_t maxlen) { 
....//strnlen might not be defined in string.h of 32-bit window's MinGW compiler!
....for(size_t i = 0; i < maxlen; i++) {
........if(s[i] == '\0') return i;
....}
....return maxlen;
}
but in the end another fatal error "_interlockedxxxxxx not defined (in boost files)" occur, which is a fatal blow to using TDM_GCC_32-bit. So TDM_GCC_64-bit is the only solution.

File-changes & commit message have further detail. Hopefully it won't conflict with Linux users (?)

It took lots of research & CPU cycle! credit to the following:
https://github.com/rubinius/rubinius/issues/2616
http://stackoverflow.com/questions/18805155/tdm-gcc-w64-script-to-change-windres-for-32bit
http://stackoverflow.com/questions/6077414/cmake-how-to-set-the-ldflags-in-cmakelists-txt
https://wiki.gentoo.org/wiki/GCC_optimization
https://springrts.com/phpbb/viewtopic.php?f=12&t=32273
https://springrts.com/phpbb/viewtopic.php?f=12&t=31678&p=554938#p554938


gajop:
Just to repeat what I said in chat, the FASTBUILD option doesn't make much sense as far as I can see.
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#Warning-Options
It basically disables warnings.

abma:
basicly TDM_GCC isn't supported. use mingw! also these changes would break build on the buildbot.

i've added more verbose info at configure time:

https://github.com/spring/spring/commit/401e2b4d9fbcd968c5067cf1be4e2508af58850c

PR #178, opened 2015-03-24
fixed unitialised memory being used as grassmap

enetheru:
this fixes the unitialised memory being used as a grass map.


abma:
looks good to me, thanks!

PR #179, opened 2015-03-25
override minimap from mapinfo.lua

enetheru:


enetheru:
btw this doesnt fail, just for some reason travid took longer than 50 minutes as can be seen in the build logs.

PR #181, opened 2015-04-01
Expose some weapon def tags to Lua nuggets

sprunk:
* dynDamageExp
* dynDamageMin
* dynDamageRange
* dynDamageInverted
* flightTime


gajop:
Looks good. It might be best to name flighttime without camel case to be consistent with the engine.
PS: I think you can do a forced push to reduce the number of commits.

PR #183, opened 2015-04-06
added lua override for infomaps

enetheru:
added three more tags for the smf section of the mapinfo.lua file to override the type,metal and grass infomaps from greyscale images. if the file is missing or invalid dimensions a content_error exception will be thrown with an informative error message

### DIFF
--- a/rts/Map/MapInfo.cpp
+++ b/rts/Map/MapInfo.cpp
@@ -333,9 +333,9 @@ void CMapInfo::ReadSMF()
 
 	smf.grassShadingTexName = mapResTable.GetString("grassShadingTex", "");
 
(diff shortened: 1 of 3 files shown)


gajop:
As always with my useful comments: fix whitespace in rts/Map/MapInfo.h

enetheru:
aww man i knew that would get me.. ive changed my vimrc so that i dont get caught out again.

jk3064:
If you wonder why I didn't merged it yet, it's cause it adds a new dependency to spring-headless which needs to be explicitly added first.

enetheru:
ooh ok. I don't know how that works. I guessed i would run into trouble with these ones since they should be synced.

PR #184, opened 2015-04-06
Add LUA Synced read functions:

DeadnightWarrior:
Spring.GetUnitFeatureSeparation(unitID, featureID[, dist2D])
Spring.GetFeatureSeparation(featureID1, featureID2[, dist2D])

There is Spring.GetUnitSeparation(unitID1, unitID2[, dist2D]), this is the extension to features, and unit vs feature distances.

### DIFF
--- a/rts/Lua/LuaSyncedRead.cpp
+++ b/rts/Lua/LuaSyncedRead.cpp
@@ -3724,7 +3724,7 @@ int LuaSyncedRead::GetUnitFeatureSeparation(lua_State* L)
 		return 0;
 	}
 
-	float3 pos1;
+	float3 pos1 = unit->midPos;
 	if (!IsAllyUnit(L, unit)) {
 		pos1 = unit->GetErrorPos(CLuaHandle::GetHandleReadAllyTeam(L));
 	}
--- a/rts/Lua/LuaSyncedRead.cpp
+++ b/rts/Lua/LuaSyncedRead.cpp
@@ -3725,16 +3725,13 @@ int LuaSyncedRead::GetUnitFeatureSeparation(lua_State* L)
 	}
 
 	float3 pos1;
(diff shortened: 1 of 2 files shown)


rts/Lua/LuaSyncedRead.cpp
```diff
@@ -3711,6 +3713,36 @@ int LuaSyncedRead::GetUnitSeparation(lua_State* L)
...
+		return 0;
+	}
+	CFeature* feature = ParseFeature(L, __FUNCTION__, 2);
+	if (feature == NULL  || !IsFeatureVisible(L, feature)) {
+		return 0;
+	}
+
+	float3 pos1;
+	if (!IsAllyUnit(L, unit)) {
+		pos1 = unit->GetErrorPos(CLuaHandle::GetHandleReadAllyTeam(L));
```
ashdnazg: pos1 won't be initialised if IsAllyUnit returns true.

DeadnightWarrior:
I've copy/pasted the code for this functions from similar ones already present in LUA Sync Read
If this is wrong, then https://github.com/spring/spring/blob/develop/rts/Lua/LuaSyncedRead.cpp#L3688 is wrong too.

ashdnazg:
Did you test if these functions actually work in the game?
You should have met errors before the fix if you had.

I ask because there was a similar case here https://github.com/spring/spring/pull/174 that could be avoided by testing.

DeadnightWarrior:
Since Spring can't be compiled with VS2010 for quite some time, I have no way of testing it ingame before a dev build appears on Spring buildbot. Previous case was purely my fault since I oversaw a use case when testing it ingame.

ashdnazg:
What about using mingw?
https://springrts.com/wiki/Building_Spring_on_Windows

GoogleFrog:
I can build spring on windows. I followed that page.

PR #186, opened 2015-04-15
Weapon loader cleanup

ashdnazg:
Some tags didn't need copying to the weapon class, since they are never changed, and don't rely on unit.


sprunk:
Aren't they there because a way to modify these is (was?) planned?

ashdnazg:
When such a way is implemented, reverting will be incredibly easy.
Until then, there's absolutely no reason to copy 10 extra vars to each weapon instance.

PR #187, opened 2015-04-16
Tag to fix burst bug

ashdnazg:
Add burstTargetLock (salvoTargetLock in lua, default false) tag to prevent weapons from changing targets during a burst (fixes #4678)


GoogleFrog:
I do not consider this a bug.

sprunk:
Hence it defaults to off (keeping current behaviour).

GoogleFrog:
Ah good. Then this looks like a good change.

ashdnazg:
Burst will probably be dealt with differently

PR #188, opened 2015-04-19
strafe air movement improvements : fix #4515,4516

springraaar:
Unit.cpp and WeaponLoader.cpp

- add lowestRange to Units and ignore weapons that are Shields when calculating it and maxRange

rts/Sim/MoveTypes/StrafeAirMoveType.cpp 

--- make flight smoother and avoid shaking  (#4515)
- When calculating intensities for rudder and aileron make them a function of the deviation from target if difference is below a threshold (value = k*deviation) instead of using +1/-1
- updateAirPhysics averages the values of rudder,aileron and elevator with the previous value when calculating the effects

--- avoid collisions with target and ground (#4516)
- make fighters abort the dive and try to climb back to normal altitude if "goalDist < (50 + min(owner->lowestRange * 0.5, turnRadius))"
- fighters start diving at 2*maxRange instead of 4*maxRange
- collision avoidance logic in GetElevatorDeflection forces aircraft to go up when below 70% of desired altitude

--- other

- override maxBank (force max 0.7) to prevent losing too much altitude when turning

- try to estimate and override turnRadius if too low (formula converts rudder, etc. from about 0.005 to 0.02 into minTurnRadius from 600 to 300)

- increased the randomness in aircraft parameters from 2% to 6% to try and get grouped aircraft to spread a bit more when moving and turning

- behavior when nearby target is overflown is only applied when attacking slow targets (ex: aircraft vs ground) and makes aircraft fly straight up to 70% turnRadius away from target before turning. Previous behavior was to pretend goal was in the opposite direction and force a turn. The new behavior allows game devs to adjust how wide the strafing patterns are by adjusting turnradius (ex: can have a highly maneuverable aircraft with a set high turn radius and long range weapons for long strafing runs).


abma:
(as already written in #sy) using multiple commits for these changes would be great!

springraaar:
ah yes, i'll remove those whitespaces. As for splitting in multiple commits i'd like to have some feedback on some of the points first

abma:
AWW, big error: please send the pull request to the develop branch. i hope you changed the develop branch? else you'll very likely get a lot of merge conflicts when updating to it.

PR #189, opened 2015-04-20
strafe air movement improvements : fix #4515,#4516 

springraaar:
Unit.cpp and WeaponLoader.cpp

- add lowestRange to Units and ignore weapons that are Shields when calculating it and maxRange

rts/Sim/MoveTypes/StrafeAirMoveType.cpp 

--- make flight smoother and avoid shaking  (#4515)
- When calculating intensities for rudder and aileron make them a function of the deviation from target if difference is below a threshold (value = k*deviation) instead of using +1/-1

- updateAirPhysics averages the values of rudder,aileron and elevator with the previous value when calculating the effects

--- avoid collisions with target and ground (#4516)
- make fighters abort the dive and try to climb back to normal altitude if "goalDist < (50 + min(owner->lowestRange * 0.5, turnRadius))"
- fighters start diving at 2*maxRange instead of 4*maxRange
- collision avoidance logic in GetElevatorDeflection forces aircraft to go up when below 70% of desired altitude

--- other

- override maxBank (force max 0.7) to prevent losing too much altitude when turning

- try to estimate and override turnRadius if too low (formula converts rudder, etc. from about 0.005 to 0.02 into minTurnRadius from 600 to 300)

- increased the randomness in aircraft parameters from 2% to 6% to try and get grouped aircraft to spread a bit more when moving and turning

- behavior when nearby target is overflown is only applied when attacking slow targets (ex: aircraft vs ground) and makes aircraft fly straight up to 70% turnRadius away from target before turning. Previous behavior was to pretend goal was in the opposite direction and force a turn. The new behavior allows game devs to adjust how wide the strafing patterns are by adjusting turnradius (ex: can have a highly maneuverable aircraft with a set high turn radius and long range weapons for long strafing runs).

### DIFF
--- a/cont/CMakeLists.txt
+++ b/cont/CMakeLists.txt
@@ -13,6 +13,7 @@ install (DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/ DESTINATION ${DATADIR}
 		PATTERN "CMakeFiles" EXCLUDE
 		PATTERN "Makefile" EXCLUDE
 		PATTERN "cmake_install.cmake" EXCLUDE
+		PATTERN "springrc-template-insource.txt" EXCLUDE
 	)
 
 ### Install freedesktop-files, mime-types etc.


jk3064:
please split at least into 3 commits

GoogleFrog:
* make fighters abort the dive and try to climb back to normal altitude if "goalDist < (50 + min(owner->lowestRange * 0.5, turnRadius))"
This looks like a fairly arbitrary formula which would depend on the sort of behaviour someone wants for their aircraft. I think this value should be a unitdef (climbDistance?) with your formula as a default.

 * fighters start diving at 2maxRange instead of 4maxRange
Looks like a major change and I have not noticed a problem with the diving behaviour in ZK with the latest engine. The range at which fighters start diving also sounds like something which should be configurable.

 * collision avoidance logic in GetElevatorDeflection forces aircraft to go up when below 70% of desired altitude
What is the desired elevation? I want fighters to fly quite low when ground attacking so perhaps this is something to configure as well?

 * override maxBank (force max 0.7) to prevent losing too much altitude when turning
Why? When messing with aircraft physics parameters it quickly becomes clear that large maxBank causes strange behaviour so people do not need protecting from entering large values. I can imagine applications for high maxBank such as a zero-gravity plane or a plane.

* try to estimate and override turnRadius if too low (formula converts rudder, etc. from about 0.005 to 0.02 into minTurnRadius from 600 to 300)
Could your formula be the default instead of an override. I like to be able to tell planes when they should think that they are able to turn around.

springraaar:
I'll try splitting into 3 commits, but could use more feedback on which changes should be split in a way that makes sense (or you want to avoid :) ). One can have the necessary changes for unit->lowestRange and most relevant part of the fix for #4516, the "abort dive when below 50% range( or turnRadius)".

I agree some of these should be unit defs, but as devs seemed against adding more I tried to do something that fit existing games as-is without adding new defs or forcing devs to change values (mostly).

Even with this change the typical fighter (500-600 range weapon) still dives dangerously close to the target. It does lower fighter dps vs ground targets though unless they use burst weapons.

About the diving at 2x maxRange instead of 4x. The previous value sort of worked, but didn't make sense. That's 2000 range for the average fighter! It also made aircraft that engage far targets fly lower for a longer period of time, subjecting themselves to collisions and ground fire.

About desired elevation, it is modified by aircraft's cruisealt. It's tipically 150-200 above the target. Maybe 70% is too strict. Without it though, aircraft avoiding collisions with each other would still bump into the ground if you have a dozen attacking a ground target. This creates a feedback that keeps them higher.

About maxBank, it isn't clear enough. XTA bombers have 0.9.

The turnRadius is only overriden if it is lower than the minimum calculated. The problem with it is that sometimes aircraft are set higher maneuverability on one direction than on others (ex: maxElevator=0.01, maxRudder=0.003), so the formula gives a minimum turnradius that's too big for them, so they stray too far from the target before turning back.

I'll try ZK. What are the unitnames of fighters and bombers there so i can spawn some quickly?

I never added any unitdef before (i'm new to this). Which do you find most relevant? Maybe the "dive safety range" ? Or the "runAfterOverflyNearbyTarget range"?

abma:
> I'll try splitting into 3 commits, but could use more feedback on which changes should be split in a way that makes sense (or you want to avoid :) 

- every commit should compile + work
- diff should be small each commit

so: every point you wrote should/could be a commit, 3 sounds very few for these changes.

PR #190, opened 2015-04-25
Add UnitStunned call-in

sprunk:
UnitStunned(unitID, unitDefID, unitTeam, stunned)
Called whenever a unit's stun status changes (ie. gets stunned, or stun wears off)


GoogleFrog:
Is UnitDamaged insufficient?

sprunk:
Currently, knowing when a unit's stun status wears off requires polling each EMPed unit.
That is sufficient but polling is worse from a performance standpoint.

PR #193, opened 2015-05-01
miscellaneous aircraft-related improvements

springraaar:
---  411068f, replaced by c0cd385

~~increase fighter/bomber aircraft maneuverability parameter randomness to 6% from 2% to spread them out a bit more when moving and turning~~

- add airManeuverabilitySpread unitDef : fighter/bomber aircraft maneuverability parameter randomness (previously hardcoded to 0.02, which is the new default)

--- 1196388
- fix fighter/bomber aircraft shaking: When calculating intensities for rudder and aileron make them a function of the deviation from target if difference is below a threshold (value = k*deviation) instead of using +1/-1

--- ad3b7d8

- added attackOverflyDistance unitDef : controls how far fighter/bomber aircraft move after overflying target. Default value of 0 is converted internally to max(turnRadius, maxWeaponRange). Only applies if attacking and against targets moving at less than 25% of the aircraft's speed.

note: previous behaviors after overflying target before turning were flying a distance function of the turnRadius (which would return 1000 for aircraft flying at any reasonable speed - too much), waiting 3 seconds after the last shot fired, and pretending a nearby goal was in the opposite side of the plane to make it turn in the opposite direction. It was meant to prevent aircraft from flying in tight circles around targets but didn't always work (ex: spawn 40 zk "fighter" and order them to attack ground, sometimes a few get stuck in tight circle and end up firing less than they could). It also calculated aileron and rudder intensity but then multiplied them by 0 to force flying straight, which seems wasteful.

--- 322d515

- randomize fighter/bomber turn direction after overflying target if it is behind in (almost) the same direction (after applying rudder in a direction, it's kept)


PicassoCT:
Is this behaviour configurable? Assuming later someone wants to write a Aircraft behaviour AI with a Gui attached to it- would they be able to undo this change?
Aircrafts are problematic anyway - changing there behaviour to be more efficen basically is a nerfing of flak. Any other gamedevs comenting on this?

springraaar:
Nope, not configurable. Previously wasn't configurable as well.

This only slightly affects acceleration and turn rates, not top speed. How would this affect someone wanting to make aircraft behaviour AI? Unless it's for a racing game...hmmmm.....

The change is slight, they still clump together more than they should. But yes, fixing aircraft behavior has balance implications. Ideally (imo) they should spread out more and avoid getting too close to the ground more.

abma:
imo: as these are "magics" the values should be configureable...

springraaar:
maybe use a unitdef, like "randomizeAirParameters?", default true, and if false would skip those instructions.

silentwings:
I don't think changing magic constants (especially ones that affect balance) is a good idea, when there is no large consensus coming out wanting it. For as long someone is silent or not interested, they are most likely to be happy with current behavior and not want it changed. 

So imo also, if you want this, it should be configurable & should default to the current value.

springraaar:
how about randomizeAirManeuverability being a float unitdef, default 2.0 (current 2% value).

game devs could then pick the value they want.

abma:
to reproduce current issue:

start zk:
/cheat
/give corhurc2

unit "shakes" when lifting off.

abma:
> maybe use a unitdef, like "randomizeAirParameters?", default true, and if false would skip those instructions.

no, i would allow to set the "aircraft maneuverability parameter" with a float/int value. which allows to directly change/set 2%/6%

springraaar:
added commits with more changes.

GoogleFrog:
Aircraft are usually uncontrollable for 5ish seconds after takeoff. This can cause them to fly into AA or just make them look really unresponsive. The behaviour is really annoying.

I made a workaround for uncontrollably in the form of this script: 
https://github.com/ZeroK-RTS/Zero-K/blob/master/scripts/fixedwingTakeOff.lua
It is spawned in script.StopMoving().

So the unitdef of corhurc2 does not tell the whole story.

springraaar:
that workaround seems to lower height to make planes exit the takeoff state sooner, but it doesn't seem like something that could break with the changes i made.

What i find though is that corhurc2 doesn't drop bombs at all on 98.0+

springraaar:
So, did anyone test this? Was it merged into the main dev branch?

One thing i didn't do is add the new CStrafeAirMoveType attributes (attackSafetyDistance, attackOverflyDistance, airManeuverabilitySpread ) to the 
"CR_REG_METADATA(CStrafeAirMoveType, (.." and "CStrafeAirMoveType::SetMemberValue(){...}", as not all the previous ones were.

abma:
no, it  wasn't merged, because:

6 changed files with 151 additions and 140 deletions. 

thats a HUGE change in vital parts of the engine. please split in smaller chunks which makes it possible to test / review the changes.

also we are pretty close to a release, thats a bad time to merge possible breaking commits.

springraaar:
I already split this in multiple commits and tested after each one, as requested. These "large" number of lines changed is mostly on a single file.

I know the release is close, but that means the next release is probably many months away.

Just add it in the next RC, we'll see if anyone complains. I bet most won't even notice unless game devs actually use the new unitdefs.

springraaar:
i see you published another RC today. And there are new commits in the dev branch, i'll merge them in too.

springraaar:
it seems to be working. Theres an issue where bombers shoot bombs too early, but it's already mantised.
https://springrts.com/mantis/view.php?id=4790

There's still that annoying camera where it starts rotating if we move the pointer top left or right....

springraaar:
So, what do we do about this pull request. Can this "migrate" to the next version?

Now that it's not going to be merged, I can make try to tweak GetElevatorDeflection to get rid of the remaning bits of shaking.

ashdnazg:
Yes, this will be dealt with in the next version.

12345swordy:
There is no reason why we should let this pull request be still up. ashnazg has done the hardcoded changes already.

springraaar:
At this point i'm not sure what was and wasn't done in later versions 101 and 102. When people start adopting Spring 102 or later, I may check aircraft behavior and make a new pull request with suggestions...or not.

I'll close it.

ashdnazg:
most of the changes weren't incorporated afaik

MaDDoXbr:
This should've been added ages ago. Why the fear of breaking something absolutely broken? O_o'

PR #194, opened 2015-05-04
fixed maxRange not being updated when UpdateRange was called

BrainDamage:
fixed maxRange not being updated when UpdateRange was called ( eg from lua's SetUnitWeaponState ) producing a series of problems such as ( but not limited to ):
- fight being oblivious to the new maxRange to seek for an enemy ( if new range was greater, fight would still use old limited maxRange and ram into enemies it could easily outrange )
- range display

### DIFF
--- a/rts/Sim/Weapons/Weapon.cpp
+++ b/rts/Sim/Weapons/Weapon.cpp
@@ -1203,6 +1203,20 @@ float CWeapon::GetRange2D(float yDiff) const
 	return (root1 > 0.0f) ? math::sqrt(root1) : 0.0f;
 }
 
+void CWeapon::UpdateRange(float val)
+{
+	float previousVal = range;
+	range = val;
+	if (val >= owner->maxRange) {
+		//we can directly assign new range
+		owner->maxRange = val;
+	} else if (previousVal == owner->maxRange) {
+		//we're changing a weapon which has same range of unit's longest weapon range
+		//new unit's range can be different, recalculate
+		owner->RecalculateMaxRange();
+	}
+}
+
 
 void CWeapon::StopAttackingAllyTeam(int ally)
 {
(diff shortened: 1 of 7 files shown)


GoogleFrog:
I worry that this will break my gunship strafe range hax. It is very useful to be able to set the range that a unit thinks it has independently of any of its weapon range. This is done by setting the max range in the weapondef and changing max range upon unit creation. Add a a unitdef which can be used to override the automatically calculated and updated max range of a unit.

Hacky gadget: https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Gadgets/unit_control_gunship_stafe_range.lua

Feature request for max range split: https://springrts.com/mantis/view.php?id=2955

rts/Sim/Weapons/Weapon.cpp
```diff
@@ -1203,6 +1203,20 @@ float CWeapon::GetRange2D(float yDiff) const
 	return (root1 > 0.0f) ? math::sqrt(root1) : 0.0f;
 }
 
+void CWeapon::UpdateRange(float val)
+{
+	float previousVal = range;
+	range = val;
+	if (val >= owner->maxRange) {
```
abma: what if the weapon with the max range gets a smaller value? can this happen?

BrainDamage: it can happen, and in that case the else if below would trigger forcing a recalculation of the max range iteratively

abma: damnit, sorry it was late :-|

thanks for explaining!

ashdnazg:
https://github.com/spring/spring/commit/37f6f0b1cd89284a1842eeb34aeba0955fce9be6

[01:33:46]  <BrainDamage> feel free to close my pull request now btw, it's useless since the hack preserving solution is in place

PR #196, opened 2015-05-10
Add a new {Game, Team, Unit}RulesParam publicity level: 'hidden'.

sprunk:
It can only be read with full read access (i.e. by LuaRules and spectators, but not any ingame players).

### DIFF
--- a/doc/changelog.txt
+++ b/doc/changelog.txt
@@ -82,6 +82,7 @@ Lua:
  - add new call-in: UnitStunned(unitID, unitDefID, unitTeam, stunned) -- called whenever a unit changes its stun status
  - add Set/GetProjectileIsIntercepted
  - add GetProjectileTimeToLive, GetProjectileOwnerID and GetProjectileTeamID.
+ - add new {Game, Team, Unit}RulesParam publicity level: 'hidden' - can only be read with full read access
  ! remove Spring.UpdateInfoTexture
  ! fix Spring.GetKeyState & Spring.PressedKeys expecting SDL2 keycodes while whole lua gets SDL1 ones
  ! Spring.PressedKeys now also returns keynames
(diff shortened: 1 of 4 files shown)


jk3064:
why isn't private enough?

sprunk:
Sometimes you want to mark stuff in a way that is not visible to affected players but still allow specs to see. Private is still readable if you are the affected team / unit owner.

jk3064:
"sometimes" ... examples?

sprunk:
Example: at some point in the game a large gaia boss unit spawns, and players shouldn't know when exactly or what unit type exactly. Hidden GameRulesParams describing the type and time allow specs to know beforehand (to build anticipation) and other gadgets can use that as well, but players won't know.

jk3064:
gaia team has rules, too?
when when those are private ...?

sprunk:
Ok: private Gaia TeamRulesParam == hidden GameRulesParam.

What about Team and Unit params?

Anarchid:
I can construct more plausible examples: a hypothetical picasso weapon that is super duper stealthy and plants eggs inside enemy units with 10% probability and neither the player using the weapon nor the player being infested should know which units are infested to build paranoia.

sprunk:
Parasite Detected

jk3064:
1. needs to be updated to latest develop
2. there are 2 logic changes (which have nothing to do with the new hidden params), those won't get merged

sprunk:
Should the game->IsGameOver() check allow to read HIDDEN (as the new max) or only up to PRIVATE (keeping previous behavior)?

Also:
(CLuaHandle::GetHandleReadAllyTeam(L) < 0) && CLuaHandle::GetHandleFullRead(L)

This checks for full read access so it won't ever be true for the ALLIED mask (even currently) because earlier PRIVATE mask checks for IsAllyUnit and that always passes with full read access (see line 365: https://github.com/sprunk/spring/blob/192ddce38f0cb69566ce277a8511b8b900107d84/rts/Lua/LuaSyncedRead.cpp#L365 ). The new HIDDEN param is also checking for full access.

PR #197, opened 2015-05-19
Add flag to draw exploded pieces recursively

ashdnazg:
if SFX.RECURSIVE is used in Explode(), the flying piece will be drawn together with its children.
(otherwise nothing is affected)


ashdnazg:
image

Silly example picture for decoration

PR #199, opened 2015-05-29
fix #4784 Improve unitsync error messages n case of duplicated files

jamerlan:
Message will be like that: 
Error: Found a "/home/revenant/.spring/maps/deltasiegedry.sd7" already in "/home/revenant/.spring/maps/DeltaSiegeDry.sd7", ignoring second one

instead of (not very helpful in my case): 
Error: Found a "tempest.sd7" already in "/home/revenant/.spring/maps/", ignoring one in "/home/revenant/.spring/maps/"

### DIFF
--- a/rts/System/FileSystem/ArchiveScanner.cpp
+++ b/rts/System/FileSystem/ArchiveScanner.cpp
@@ -566,7 +566,7 @@ void CArchiveScanner::ScanArchive(const std::string& fullName, bool doChecksum)
 			} else {
 				if (aii->second.updated) {
 					const std::string filename = aii->first;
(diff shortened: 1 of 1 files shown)


cleanrock:
I think it should say "ignoring first one".

abma:
imo it should just say "ignoring" :)

jamerlan:
Updated to "ignoring".

cleanrock:
"ignoring" is a safe bet if unsure what is ignored ;)

jamerlan:
Its not matters actually. Its a signal about duplicated file

cleanrock:
I appreciate the contribution but if we know which file is ignored it is good information. Using only "ignored" is hinting that first file is ignored so i am happy.

abma:
thanks!

PR #204, opened 2015-06-04
Los recolor for some decent defaults.

GoogleFrog:
The current default is quite bad because the settings were invalidated by the new infotex system. The previous settings looked like the ZK settings from before ZK updated its values. The new settings are the new ZK default.


abma:
thanks!

PR #207, opened 2015-06-12
Fix Lua.callRules, Lua.callUI for JavaOO

rlcevg:
[Lua<->AI communication](https://springrts.com/phpbb/viewtopic.php?f=15&t=26298) now works in 3/4-duplex (no ret_outData in SLuaMessageEvent).

Purpose of [ret_outData](https://github.com/spring/spring/blob/d1c9887f8e8c8931f5a71d73a82d05ac5c48e222/rts/ExternalAI/Interface/AISCommands.h#L412) from SCallLua(Rules|UI)Command was changed so now it points to a buffer where response should be copied, i.e. caller is responsible for providing proper buffer (or NULL).

I see 1 (dis)?advantage: buffer for ret_outData must have const size MAX_RESPONSE_SIZE=10240 chars (defined in aidefines.h). An alternative would be an additional "int outSizeMax" param in SCallLua(Rules|UI)Command. But that will require more hassle with heap memory opposed to stack in awk scripts, and such generated interface will annoy with additional param outSizeMax.

Tested Lua::CallUI with ZK + LegacyCpp, Cpp, JavaOO AIs.

### DIFF
--- a/AI/Interfaces/Java/bin/jni_wrappCallback.awk
+++ b/AI/Interfaces/Java/bin/jni_wrappCallback.awk
@@ -135,10 +135,11 @@ function printNativeJNI() {
 				pType_c   = params[p];
 				sub(/ [^ ]+$/, "", pType_c);
 				pType_c = trim(pType_c);
-				pType_jni = convertCToJNIType(pType_c);
 
 				pName = params[p];
 				sub(/^.* /, "", pName);
+
+				pType_jni = convertCToJNIType(pType_c, pName);
 
 				# these are later used for conversion
 				c_paramTypes[p] = pType_c;
@@ -173,6 +174,7 @@ function printNativeJNI() {
 				retTypeConv = 1;
 			}
 
+			hasRetParam = 0;
 			# Params conversion - pre call
 			for (p=1; p <= size_params; p++) {
 				pType_jni = jni_paramTypes[p];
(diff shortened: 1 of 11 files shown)


Anarchid:
Awesome!

rlcevg:
Maybe it is possible to do full-duplex trick with MAX_RESPONSE_SIZE for SLuaMessageEvent.
EventsJNIBridge.c should be generated with something like
```C
...
case EVENT_LUA_MESSAGE: {
  const struct SLuaMessageEvent* evt = (const struct SLuaMessageEvent*) data;
  jstring inData_jni = (*env)->NewStringUTF(env, evt->inData);
  // Create StringBuffer object (outData)
  jclass clazz = (*env)->FindClass(env, "StringBuffer");
  jobject ret_outData_jni = (*env)->AllocObject(env, clazz);
  // Execute luaMessage event
  _ret = (*env)->CallIntMethod(env, aiInstance, m_ai_luaMessage, inData_jni, ret_outData_jni);
  // Get char* from StringBuffer
  jmethodID toStringID = (*env)->GetMethodID(env, sbufClass, "toString", "()Ljava/lang/String;");
  jstring js = (jstring) (*env)->CallObjectMethod(env, jarg1, toStringID);
  const char *pCharStr = (*env)->GetStringUTFChars(env, js, 0);
  // Copy response back to engine
  size_t len = min(strlen(pCharStr), MAX_RESPONSE_SIZE - 1);
  strncpy(evt->ret_outData, pCharStr, len);
  evt->ret_outData[len] = '\0';
  // Free resources
  (*env)->ReleaseStringUTFChars(env,  js, pCharStr);
  break;
}
...
```
But engine(caller) must allocate proper response buffers.

Is full-duplex even worth the hassle?

PR #208, opened 2015-06-20
Fixed LightningCannon trouble with hiting targets.

GoogleFrog:
This fixes issues that LightningCannons have with hitting their target.The bug most obviously affects shieldfelon and armzeus from ZK.

The fix is similar to the fix for Rifle: https://github.com/spring/spring/commit/71b7fc3b7afb8132fdaf8da7b3b952110750562f


ashdnazg:
Thanks!

PR #209, opened 2015-06-24
add cmake option DEBUG_LUANAN to enable lua check for nans in a release build

abma:
thoughts about this?

i'm unsure if this change is good or bad:

pro:
- allows to test for nans in lua code with release builds (debug is horrible slow)
- only rts/lib/lua has to be recompiled when enabled

con:
- when a nan is found it insta-desyncs which maybe leads to more confusion
- compile setting (runtime setting would be nicer, but slow)
- new compile setting = more complexity

(+things i can't think about atm)

maybe the better solution is to make the used lua a shared lib so functions can be replaced with the LD_PRELOAD trick?

### DIFF
--- a/doc/changelog.txt
+++ b/doc/changelog.txt
@@ -174,7 +174,7 @@ CMake:
  ! remove DEBUG2 & DEBUG3, replaced with just DEBUG
  - enable "Identical Code Folding" with ld.gold
  - disable unused ASSIMP formats at compiletime
-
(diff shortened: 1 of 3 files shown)


ashdnazg:
I think this is a very good idea.
A game-developer oriented debug build that contains input checks (not only here, but in other places as well) will help pretty much everybody.

PR #212, opened 2015-08-03
Added dynamic Piece Hierarchy

PicassoCT:
This is not debugged yet- please do not merge.

Would be glad, if the build bot could make a windows build so that i can test it with the mingw-dbg.


abma:
please create a new pull request which contains only the changes you did.

this pull request contains A LOT of files which must not be in the git repository! (src-generated / .so / ...)

abma:
this seems partly our fault, .gitignore is incomplete. mantised: https://springrts.com/mantis/view.php?id=4922

PR #213, opened 2015-08-04
Added dynamic Piece Hierarchy - 

PicassoCT:
this time with a clean commit. Thanks go to gajop and jk


abma:
can you please remove all example code/data from rts/Sim/AdvUnitPhysics/ before creating a commit / the pull request. also external libs go into rts/lib... i assume you didn't change/create 1.6k files.

this commit contains way to much data, the spring repository is already pretty large and increasing size with unneeded data/files should be avoided.

PicassoCT:
Sorry, this not supposed to be at all in the commit

abma:
in short: don't use git add -A

add files one by one.

PR #216, opened 2015-08-05
Add support for SSMF Detail Normal Textures

Beherith:
Works similarly to the previous SSMF splatting, but instead of sampling a combined texture of 4 different single-channel diffuse detail textures, it samples four separate normal textures.
The textures must contain the normal maps in the RGB channel (standard normal maps), with a diffuse luminance in the alpha channel.
The system uses the same splatDistrTex as regular SSMF.

Usage:
Specify splatDetailNormalTex1 .. splatDetailNormalTex4 in the MAP resources section.
splatDetailTex will be unused if the above are present.
specify splatDetailNormalDiffuseAlpha=1 to enable diffuse textures stored in alpha channels

~10% performance cost of sampling 4 textures once each vs the previous 1 texture 4 times.

Screenshot: http://beherith.eat-peet.net/stuff/detail_normals.png
Example map: http://beherith.eat-peet.net/stuff/DeltaSiegeDry_Reloaded_vBETA3.sd7

Issues discussed with jK in previous pull request have been addressed. (https://github.com/spring/spring/pull/131)

### DIFF
--- a/cont/base/maphelper/maphelper/mapdefaults.lua
+++ b/cont/base/maphelper/maphelper/mapdefaults.lua
@@ -103,6 +103,11 @@ local mapDefaults = {
     detailNormalTex   = '',
     lightEmissionTex  = '',
     parallaxHeightTex = '',
+    splatDetailNormalTex1 = '',
+    splatDetailNormalTex2 = '',
+    splatDetailNormalTex3 = '',
+    splatDetailNormalTex4 = '',
+    splatDetailNormalDiffuseAlpha= false,
   },
 
   defaultTerrainType = {
--- a/cont/base/maphelper/maphelper/parse_tdf_map.lua
+++ b/cont/base/maphelper/maphelper/parse_tdf_map.lua
@@ -207,6 +207,11 @@ return function(sourceText)
      detailNormalTex   = map.detailnormaltex,
      lightEmissionTex  = map.lightemissiontex,
      parallaxHeightTex = map.parallaxheighttex,
+     splatDetailNormalTex1 = map.splatdetailnormaltex1,
+     splatDetailNormalTex2 = map.splatdetailnormaltex2,
+     splatDetailNormalTex3 = map.splatdetailnormaltex3,
+     splatDetailNormalTex4 = map.splatdetailnormaltex4,
+     splatDetailNormalDiffuseAlpha = map.splatdetailnormaldiffusealpha,
   }
 
   ConvertTerrainTypes(map)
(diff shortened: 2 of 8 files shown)


ashdnazg:
@jk3064 input?

Beherith:
@jk3064 @rtri , I have removed the duplications, is there anything else you feel that is still required?

rtri:
suspended pending http://imolarpg.dyndns.org/trac/balatest/ticket/932

abma:
the BAR issue is fixed

Beherith:
Any chance someone could look at this? I have tested it with existing maps extensively, and it causes no issues with any existing content.

gajop:
I'd like this resolved too. I have a number of changes in the same files which I'm hesitant to do so as not to create conflicts.
Still, can we get some information on how to create the four normal splat maps? I don't know how to use this.

PS: Why do we have normals as splats but diffuse as "regular" textures? Unity seems to use splat diffuse textures too (at least in their map editor as can be seen here: http://i.imgur.com/k7e2oKD.jpg). Are any of these choices justified by quality/size or are they just random?

Beherith:
The four detail texture normal splats can just be any 3-channel normal map, or they can all contain a greyscale diffuse texture in their respective alpha channels. 

DNTS is not XOR with the current global normal map (detailNormalTex, whose name is quite confusing, as it is not 'detail' level, but a single texture stretched onto the whole map), and both can be used simultaneously. 

Gajop, take a look at the posted example map archive. I would say that DNTS is not much more work to set up than regular SSMF.

I know that free time is far from free, as time is our only finite resource, so I would like to ask what I could barter in return for a developer taking a look at this. If its just as simple as me providing extensive documentation and examples on the springrts wiki, then let me know!

gajop:
Imo for merging this only some wiki documentation should be added like it was done in some other cases (example images and description)
===================

Yes the detailNormalTex is a bit confusing. I'm not actually sure which format it's stored in and rendering on that texture directly in-game was met with mixed success. 

I've checked your map/engine before and it worked fine for me. That said I don't know how to support both splat (normals, splat detail(?)), and streched/multiple (specular, diffuse) in an intuitive way for the user of a GUI map editor. Usually you would select a single material (which includes the diffuse and sometimes normal, specular, glow and height maps), and you would paint that material on the terrain.  I'm thinking about this from a tool-writer standpoint, which is perhaps a bit unrelated, but to me it's kinda confusing to have specular as stretch, normals as 4 textures done by splatting, and diffuse as multiple, fixed-size textures (why not just use a single format for all with maybe different resolution?)

Beherith:
Some documentation: https://springrts.com/wiki/Mapdev:splatdetailnormals

PR #219, opened 2015-08-23
Fix an error regarding va_start

12345swordy:


abma:
fixed slightly different, thanks!

https://github.com/spring/spring/commit/37ba0163d1a9a52900ae81ebc14be5f45cbea580

PR #223, opened 2015-11-12
Manual unit share: allow EMPed units

sprunk:
The decision to block this belongs to the mod.

### DIFF
--- a/rts/Net/NetCommands.cpp
+++ b/rts/Net/NetCommands.cpp
@@ -887,6 +887,8 @@ void CGame::ClientReadNet()
 
 						if (unit->isDead)
 							continue;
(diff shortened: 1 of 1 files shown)


ashdnazg:
What will happen if you gave a nanoframe inside a factory while keeping the factory?

sprunk:
I suspect nothing but for safety I changed the PR to only unblock EMPed unit transfer (because this is safe).

ashdnazg:
thanks!

there should probably be an "AllowGive" callin

sprunk:
There is an AllowUnitTransfer call-in already.

ashdnazg:
ah, good.

PR #225, opened 2015-11-16
Reduce shadow acne

Beherith:
The small negative Z bias moved the point closer to the light, so it gets less acne, and the clamping makes sure we are only shadowing areas that are more than 50% occluded. Ill try to post comparison screenshots if needed.


rtri:
Adding an R-value bias is fine, the PCF part is personal preference (and won't do much on hardware without linear SM filtering) best left to a custom shader.

PR #221, opened 2015-09-16
merge pr-downloader branch

abma:
thoughts about this?

this would make pr-downloader mandatory, atm its optional. is this ok or should it still be optional?

the code still needs some improvements, but the api / file structure should stay the same.

PR #228, opened 2015-11-18
add VFS.CalcMd5 function

gajop:
interested if there are any objections before merging this.
see https://github.com/spring/spring/pull/221#commitcomment-13841525 and comments below

### DIFF
--- a/doc/changelog.txt
+++ b/doc/changelog.txt
@@ -11,7 +11,7 @@ General:
 Lua:
  ! GameID callin now gets the ID string encoded in hex.
  - new Spring.GetUICommands function to obtain a list of all UI commands (e.g. /luaui reload)
- - new VFS.CalcMd5 function that calculates the MD5 hash (in base64 form) of a given string
+ - new VFS.CalculateHash function that calculates hash (in base64 form) of a given string (with md5 support initially)
 
 -- 100.0 --------------------------------------------------------
 Major:
--- a/doc/changelog.txt
+++ b/doc/changelog.txt
@@ -6,10 +6,12 @@ Spring changelog
 
 General:
  ! change default screenshot file type to jpg. to create png use /screenshot png
+ - pr-downloader dependency is now mandatory (statically linked with spring)
 
 Lua:
  ! GameID callin now gets the ID string encoded in hex.
- - new GetUICommands function to obtain a list of all UI commands (e.g. /luaui reload)
+ - new Spring.GetUICommands function to obtain a list of all UI commands (e.g. /luaui reload)
+ - new VFS.CalcMd5 function that calculates the MD5 hash (in base64 form) of a given string
 
 -- 100.0 --------------------------------------------------------
 Major:
--- a/rts/Lua/LuaVFS.cpp
+++ b/rts/Lua/LuaVFS.cpp
@@ -19,6 +19,8 @@
 #include "System/FileSystem/VFSHandler.h"
 #include "System/FileSystem/FileSystem.h"
 #include "System/Util.h"
+#include "../tools/pr-downloader/src/lib/md5/md5.h"
+#include "../tools/pr-downloader/src/lib/base64/base64.h"
 
 using std::min;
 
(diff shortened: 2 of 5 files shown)


gajop:
@jk3064 thanks, updated

rtri:
Even if you wave your hands and pretend Lua code might at some point have a legitimate reason to calculate file hashes (instead of the real use-case which is to process legacy lobby passwords) so that this function has a right to belong in LuaVFS, it is inflexible because only one type of hash is supported.

Since continued promotion of and dependence on MD5 is a bad idea but the downloader lib is tied to it already for pool archives, the responsibility of calculating the hash should not be placed inside LuaVFS but moved to prdl instead. The *only* lines added here should be...

    int LuaVFS::CalculateHash(lua_State* L) {
        lua_pushsstring(L, prdl::CalculateHash(luaL_checksstring(L, 1), luaL_checkint(L, 2)));
        return 1;
    }

... where the second argument denotes hash-type mapped by some enum (0=MD5, 1=SHA1, etc), so that pr-downloader can transparently implement whichever are needed.

gajop:
@rtri This main intention for this is indeed to provide the login functionality for the lobby server. I make no other claims, and seeing as adding the relatively complex crypto stack needed for secure logins seems difficult and due to the also disabled no-password login, this is a reasonable thing to do.

While I don't think SHA hash passwords are any safer, I agree that this could all be done in the pr-downloader (or alternative library), and thus be reusable in other parts of the engine for whatever reason. I'll make the suggested changes.

rtri:
Client-side password hashing is obviously insane, and supporting hashes other than MD5 should not be taken as an endorsement of that idea or regarded as improving safety.

Again, just pretend a need exists for a Lua lobby to hash file contents using (a) more advanced method(s) of its choice.

wrt linking, a small utility library that both Spring and pr-downloader could depend on has my preference.

abma:
> wrt linking, a small utility library that both Spring and pr-downloader could depend on has my preference.

does such a "small" lib already exist? if not a candidate is openssl/libcrypto which is far away from beeing small imo. an alternative could be nettle: http://www.lysator.liu.se/~nisse/nettle/

abma:
hm, maybe lsl could be used?! : https://github.com/springlobby/lsl

but it depends on boost, jsoncpp and  contains cimg, so isn't small, too (but includes md5 + base64)

gajop:
For now I suggest we just use what we already have as to not complicate things.
A third party library could likely be a part of the security stack we may add to support the safe protocol, so it would probably be best to pick one for both use cases.

abma:
> For now I suggest we just use what we already have as to not complicate things.

i agree with that as i don't see better alternatives :-|

only thing we could do is to keep pr-downloader optional by checking in cmake if the directory exists and then set a compiler flag to disable the code.

ashdnazg:
what's the reasoning behind making pr-downloader optional?

abma:
> what's the reasoning behind making pr-downloader optional?

atm it is optional and pr-downloader pulls curl as dependency which depends on a lot of other libs, so nothing big, more a wish.

gajop:
I don't mind it being optional, as long as it's enabled by default.
No idea how to write that code though.

gajop:
With that in mind, how about I move this into pr-downloader and merge it, you can then make it an optional dependency?

gajop:
This should be it after abma merges the pr-downloader PR: https://github.com/spring/pr-downloader/pull/84
I had to make the API a bit more complicated so it:
- less error prone in case of different encryption types
- works using the C-API (C strings)

ashdnazg:
You need to merge the pr-downloader changes from your repo to the main

gajop:
I don't have push privileges to spring/pr-downloader (or spring/ubserver
for that matter), only spring/spring.
Hence why I'm waiting for someone (abma?) to merge it

2015-11-22 0:59 GMT+09:00 Eshed <notifications@github.com>:

rtri:
IMO still too many lines.

The lib "knows" which hashtypes are valid, LuaVFS does not. Make CalcHash return a nullptr (or empty string so std::string(...) does not crash) for invalid values.

gajop:
There you go

2015-11-22 1:03 GMT+09:00 rtri <notifications@github.com>:

gajop:
I think it's good enough now so I've merged it. At least there shouldn't be any dramatic API changes.
If you feel there's a need to do any further modifications let me know.

PR #230, opened 2015-11-24
Feature defs: add a lockPosition tag to control XZ movement lock

sprunk:
Controls whether the feature is locked to an XZ position, like currently trees, rocks etc are.
Value is an int: -1 is current behaviour (locked iff not resurrectable), 0 is non-locked, 1 and everything else is locked

Also:
* non-locked features now obey the "floating" tag.
* fixes non-locked features having double vertical velocity (speed was already applied in Feature.cpp:519 but vertical was applied again in 540)
* fixes locked features keeping residual horizontal impulse (which caused flickering when it was high)

### DIFF
--- a/doc/changelog.txt
+++ b/doc/changelog.txt
@@ -8,6 +8,10 @@ General:
  ! change default screenshot file type to jpg. to create png use /screenshot png
  - pr-downloader dependency is now mandatory (statically linked with spring)
 
+ FeatureDefs:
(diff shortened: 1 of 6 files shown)


ashdnazg:
I'd have put it in the feature itself and not the def (maybe a default in the def).
Then you can change it from lua.

sprunk:
see #231

PR #231, opened 2015-11-25
Feature XZ position lock patch

sprunk:
Features can currently have their XZ position locked (only able to move along the Y vector) but this is hardcoded and tied to resurrectability (locked iff not resurrectable). The patch adds a way to control that and fixes some related issues.

* add Spring.{Get,Set}FeatureLockPosition Lua functions to control the XZ position lock.
* non-locked features now obey the "floating" tag.
* fix locked features keeping horizontal impulse (which caused rendering flickering when it was high).
* fix non-locked features having double vertical velocity.
* fix features not inheriting velocity if created from other features.

### DIFF
--- a/rts/Lua/LuaSyncedCtrl.cpp
+++ b/rts/Lua/LuaSyncedCtrl.cpp
@@ -231,6 +231,7 @@ bool LuaSyncedCtrl::PushEntries(lua_State* L)
 	REGISTER_LUA_CFUNC(SetFeatureMidAndAimPos);
 	REGISTER_LUA_CFUNC(SetFeatureRadiusAndHeight);
 	REGISTER_LUA_CFUNC(SetFeatureCollisionVolumeData);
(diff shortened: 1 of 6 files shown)


sprunk:
Tested, and any problems fixed.

sprunk:
I'm now unsure about 5), it probably needs a bit more consideration. The case I'm thinking of is trees and upward impulse. Current behaviour doesn't let trees jump up from impulse and I think this is the wanted behaviour so I'd like to keep it. Allowing Lua to override this would be good but it would require some way to differentiate between speeds from impulse and Lua.

rtri:
The answer is FeatureMoveCtrl.

I won't apply this PR, but should have an API ready soonish.

ashdnazg:
@rtri partially, the check for udef != NULL can be replaced with something more flexible which requires less work then a FeatureMoveCtrl implementation.

PR #232, opened 2015-11-30
Allow SetUnitVelocity in UnitDestroyed

sprunk:
The patch makes wrecks save the parent unit's speed from after UnitDestroyed (currently is before). This allows to set wrecks' initial speed reliably.


ashdnazg:
thanks :)

PR #233, opened 2015-12-02
Add a new modrule: experience.lastHitMult

sprunk:
A multiplier for the last hit XP bonus. Its default value is 1 which is the current behaviour, ie. the kill bonus is the same as bringing that unit from 100% to 0% health.


sprunk:
It appears that AddUnitExperience to offset the bonus with a negative amount should suffice. Will reopen if any problems with that call show up.

PR #234, opened 2015-12-11
Add a field "default" to the list of known widgets, so that an indivi…

springjools:
…dual widget can access a list of widgets that are enabled/disabled by default.


ashdnazg:
bad indentation.
Also what's the motivation?

PR #237, opened 2015-12-20
add ability to set map shading textures

gajop:
This should add some basic ability to modify engine textures. I've tested with "ssmf_specular" and it works: http://imgur.com/gWrECvM . That said, the code is still WIP and I'll accept comments on how it can be improved (personally I think adding functions to list/enable/disable ssmfs would be useful).

Follow up on the convo (https://github.com/spring/spring/commit/05dce347dcdfc37d3ed5f602ffc2b5b86365b12f#commitcomment-15061940) I had with @rtri after he exposed more engine textures to read: 
I think having the engine track Lua textures isn't problematic. It's rather useful in the sense that you get a WYSIWYG editor, and having the shading texture equivalent for SetMapSquareTexture is useful and keeps things consistent. The ability to enable/disable certain shading textures can also be added as a callout, and it makes the engine more flexible I think.

Thoughts?


rtri:
I'll commit a cleaned/fixed version without the copypasta in a minute.

gajop:
Cool, thanks!

PR #242, opened 2016-02-09
Update parse_tdf_map.lua to support Detail normal texture splatting in .SMD files 1/2

Beherith:


rtri:
I think this is a good point to start letting SMD die actually...

abma:
> IMO this is a good point to start letting SMD die, no new map should still use the tdf format.

i agree to this, but IMO we need a way to deprecate maps, so these can be replaced with a fixed version.

-> maps using tdf needs some deprecated warning

does a conversation utility exist for this?

abma:
-> https://springrts.com/mantis/view.php?id=5189

PR #246, opened 2016-02-20
add float3.cpp to AI/Wrappers/Cpp lib

cleanrock:
AIFloat3 inherits float3.


cleanrock:
CircuitAI did not run because float3::equals was not found since its no longer inline, float3.cpp should be included in libCpp.a since AIFloat3 inherits float3.

ashdnazg:
thanks :)

PR #248, opened 2016-03-24
add '\GroundDetail <value>'

silentwings:
(+ minor consistency between SM3 and SMF)


abma:
why backslash? all commands are afaik prefixed with a slash?!

also in several places spaces are used when tabs should be used. 

general style hint:

+		int detail;
+		if (!(action.GetArgs()).empty()) {
+			detail = atoi((action.GetArgs()).c_str());
+		} else {
+			LOG_L(L_WARNING, "/%s: missing argument", GetCommand().c_str());
+			return false;
+		}

vs

+		if (action.GetArgs().empty()) {
+			LOG_L(L_WARNING, "/%s: missing argument", GetCommand().c_str());
+			return false;
+		}
+		const int detail = atoi((action.GetArgs()).c_str());


idk whats better:
1. add a new command
2. add a hook to SetConfigInt which updates the runtime value


atm i would prefer 2. but idk if there are major drawbacks with this approach.

nicfreeman1209:
> why backslash? all commands are afaik prefixed with a slash?!
Because thats what every single other action executor does when its given the wrong syntax. 

> 2. add a ConfigObserver which updates the runtime value
This would not match the behaviour of any other GUI related spring setting that I have ever come across. The solution above matches the current behaviour of the vast majority of spring settings that offer UI commands.

> general style hint:
Again, I have simply done what the majority of other action executors in the that very same file do. I'm not changing them all.

>  in several places spaces are used when tabs should be used. 
This is the only point that seems valid to me; I'd be happy to change them but frankly if you're going to be (wrongly, imo, as said) picky about all the above points them I'm not going to bother.

> basicly you should only a config value once.
I don't really see what this has to do with it, since it *is* only set once, since it costs nothing to change the value anyway. In any case, Spring.SetConfigInt could get called by the user as often as they fancy.

nicfreeman1209:
Could this recieve a response?

It is not clear to me if it has been
(1) outright rejected on philosophilca grounds or
(2) would be accepted if tab/whitespace was fixed.

ashdnazg:
doesn't seem like this is going to break anything, although there should be a real solution somewhere for all these very similar problems.
Fix the style/whitespace and I'll merge

nicfreeman1209:
Done.

nicfreeman1209:
Actually, it would probably be cleaner if I just sent it all as a single commit in a new pull req.

nicfreeman1209:
Cleaned up into single commit in https://github.com/spring/spring/pull/250

PR #249, opened 2016-04-03
Fix grass recieving shadows

Beherith:
On my nvidia card, shadow2DProj().a always returned 1. Also, clamping the value + groundshadowdensity was incorrect imo, as the greater the density, the less of an effect it had because of the addition.


rtri:
Seems the same logic is in TreeFragProg, which differs from model / terrain / decal shadowing.

Your fix is more sensible, but I would prefer consistent application of groundShadowDensity across all shaders [i.e. `1.0 - (1.0 - shadowCoeff) * groundShadowDensity` == `mix(1.0, shadowCoeff, groundShadowDensity)`] to a third variant.

Beherith:
I agree with your consistent application, the difference is minor,

PR #251, opened 2016-04-15
Added StructurePlacementNotify Lua unsynced callin

lhog:
This is to allow for native drawing of blocked squares, defined by mod. Say you, as a mod builder, would like to:

-  disallow building of any type except metal extractors on metal spots
-  or you want metal extractors to be only buildable on certain location
-  or you don't want non geo-enabled buildings to occupy geo spot.

Existing solutions can track such construction requests using CommandNotify in unsynced context. Few issues exist however:

1. Blocked squares cannot be drawn in a native way (i.e. as yellow/reg/green grid under the building blueprint). DrawWorld call-in is called before mentioned grid and building's model are drawn when you hover mouse with active command set. Therefore whatever you have drawn in DrawWorld becomes overwritten afterwards.

2. Complex construction modes like, line build, rectangular build are not easy to track from Lua: It should be done before the command is issued (while you're moving your mouse and setting up blueprint for future line/rect construction), and therefore should give appropriate visual feedback. This requires replicating of lots of stuff, done by engine anyway and taking into account quite a few events such as keys pressed (shift, alt, ctrl), mouse move, mouse button clicks, build spacing, selected mode (single building, line, rectangle, hollow rectangle).

The proposed solution augments **unsynced** part of CGameHelper::TestUnitBuildSquare with optional Lua callin, which asks Lua mod if particular building can be built in particular place. It also requests for mod-blocked squares, this information is then merged with engine-blocked squares (like another building is there, slope is too steep, water, etc) and displayed in the same way. Handling of such callin is optional on mod side, if it's not defined in the Lua, then nothing will change.

I didn't touch synced part on purpose as AllowCommand or AllowUnitCreation could be used to do same already for synced context (AllowUnitCreation parameters are fact is very similar to parameters, supplied to the proposed call-in). The reason I didn't touch sync was that I didn't want to multiply controlling entities. If requested, the code could be modified, such as that a synced part is handled as well. At least I think It's possible.

### DIFF
--- a/cont/LuaUI/callins.lua
+++ b/cont/LuaUI/callins.lua
@@ -17,6 +17,8 @@ CallInsList = {
   "LayoutButtons",
   "ConfigureLayout",
   "CommandNotify",
+  
+  "StructurePlacementNotify",
 
   "KeyPress",
   "KeyRelease",
--- a/cont/LuaUI/main.lua
+++ b/cont/LuaUI/main.lua
@@ -100,6 +100,10 @@ function CommandNotify(id, params, options)
   return widgetHandler:CommandNotify(id, params, options)
 end
 
+function StructurePlacementNotify(unitDefID, x, y, z, buildHeight, x1, z1, x2, z2, allyteam)
+  return widgetHandler:StructurePlacementNotify(unitDefID, x, y, z, buildHeight, x1, z1, x2, z2, allyteam)
+end
+
 function DrawScreen(vsx, vsy)
   widgetHandler:SetViewSize(vsx, vsy)
   return widgetHandler:DrawScreen()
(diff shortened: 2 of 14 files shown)


jk3064:
you are mixing synced & unsynced here (GameHelper.cpp contains synced code) :/

lhog:
@jk3064, saw your comment in the email notification, but for some reason can't find it here on GitHub, not sure why.

Anyway I think the [other solution](https://github.com/spring/spring/pull/259) to the issues described on this page is better as:
* it's one or few times push (likely covering only several tiles, like mex or geo locations) from Lua vs. constant pull to Lua side.
* It works on both synced and unsynced side seamlessly.

PR #252, opened 2016-04-16
Applied decloakUnitShape attribute to unit (was defined only on UnitDef level)

lhog:
Proposed for the purpose of being able to change decloakUnitShape attribute on the fly through Lua synced control. This can be useful if cloaked unit changes its realm. For instance decloak may be spherical for a unit on the ground, but change to cylindrical when the unit is in the water.

### DIFF
--- a/doc/changelog.txt
+++ b/doc/changelog.txt
@@ -24,6 +24,7 @@ Lua:
  - Add Spring.GetFrameTimer(bool synced) to get a timer for the start of the frame
    this should give better results for camera interpolations.
  - Add an optional weaponNum argument to SetUnitTarget.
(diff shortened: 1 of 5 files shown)


ashdnazg:
what's a use case?

lhog:
@ashdnazg, it's outlined in the PR comment more or less. Mostly it's for cloaked amphibian units. They should have sphere decloaking shape on land and cylinder under water (otherwise ship or hover units may fail to decloak).

Anarchid:
"It is needed for sonar to die"

lhog:
Perhaps I should better propose yet another Lua call-in. Something like AllowCloakStateChange. @ashdnazg is right that cylindrical decloak shape can make aircrafts overpowered vs cloaked submerged amphibians.

PR #253, opened 2016-04-17
Added AllowProximityDecloak synced callin

lhog:
Synced AllowProximityDecloak is here to possibly cancel engine inducted decloak of a cloaked unit, entered its decloak radius. The use case is to be able to better control when a unit is decloaked from the mod. One example is amphibian cloaked unit, which is supposed to have sphere decloak shape on land and cylinder shape in the water. Though cylinder height should be limited to some value a bit above water level (otherwise air will be able to decloak amphibian in the water). So in order to achieve desired behavior, the decloakShape will be set  to cylinder in UnitDef and then checked inside AllowProximityDecloak whether the decloaking unit is within sphere if cloaked unit is on land or whether the decloaking unit is within height-limited cylinder when cloaked unit is under water.

Proposed as a better alternative to https://github.com/spring/spring/pull/252

### DIFF
--- a/cont/base/springcontent/LuaGadgets/callins.lua
+++ b/cont/base/springcontent/LuaGadgets/callins.lua
@@ -104,6 +104,7 @@ CALLIN_LIST = {
 	"AllowResourceTransfer",
 	"AllowDirectUnitControl",
 	"AllowBuilderHoldFire",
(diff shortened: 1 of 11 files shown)


lhog:
Any movement here or formal feedback?

PR #254, opened 2016-04-19
Remove license text from loadscreen, add to README.markdown and infolog

sprunk:
Edit: the text now lives in README.markdown and is removed from loadscreen completely
Edit2: also lives in infolog

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -7,6 +7,10 @@ Spring (formerly TASpring) is an Open Source Real Time Strategy game engine.
 Visit our [project homepage](http://springrts.com/) for help, suggestions,
 bugs, community forum and everything spring related.
 
+## License
+
+This program is distributed under the GNU General Public License, see license.html for more info
+
 ### Compiling
 
 Detailed instructions for how to compile Spring can be found [here](http://springrts.com/wiki/Building_spring)
--- a/rts/Game/LoadScreen.cpp
+++ b/rts/Game/LoadScreen.cpp
@@ -42,6 +42,7 @@
 
 CONFIG(int, LoadingMT).defaultValue(-1).safemodeValue(0);
 CONFIG(bool, ShowLoadMessages).defaultValue(true);
+CONFIG(bool, LoadscreenLicenseLeftAlign).defaultValue(false);
 
 CLoadScreen* CLoadScreen::singleton = nullptr;
 
@@ -333,10 +334,18 @@ bool CLoadScreen::Draw()
 	font->Begin();
 		font->SetOutlineColor(0.0f,0.0f,0.0f,0.65f);
 		font->SetTextColor(1.0f,1.0f,1.0f,1.0f);
-		font->glFormat(0.5f,0.06f, globalRendering->viewSizeY / 35.0f, FONT_OUTLINE | FONT_CENTER | FONT_NORM,
-				"Spring %s", SpringVersion::GetFull().c_str());
-		font->glFormat(0.5f,0.02f, globalRendering->viewSizeY / 50.0f, FONT_OUTLINE | FONT_CENTER | FONT_NORM,
-			"This program is distributed under the GNU General Public License, see license.html for more info");
+		bool leftAlign = configHandler->GetBool("LoadscreenLicenseLeftAlign");
+		if (leftAlign) {
+			font->glFormat(0.02f,0.06f, globalRendering->viewSizeY / 35.0f, FONT_OUTLINE | FONT_NORM,
+					"Spring %s", SpringVersion::GetFull().c_str());
+			font->glFormat(0.02f,0.02f, globalRendering->viewSizeY / 50.0f, FONT_OUTLINE | FONT_NORM,
+				"This program is distributed under the GNU General Public License, see license.html for more info");
+		} else {
+			font->glFormat(0.5f,0.06f, globalRendering->viewSizeY / 35.0f, FONT_OUTLINE | FONT_CENTER | FONT_NORM,
+					"Spring %s", SpringVersion::GetFull().c_str());
+			font->glFormat(0.5f,0.02f, globalRendering->viewSizeY / 50.0f, FONT_OUTLINE | FONT_CENTER | FONT_NORM,
+				"This program is distributed under the GNU General Public License, see license.html for more info");
+		}
 	font->End();
 
 	if (!mtLoading)
--- a/rts/Game/LoadScreen.cpp
+++ b/rts/Game/LoadScreen.cpp
@@ -42,7 +42,6 @@
 
 CONFIG(int, LoadingMT).defaultValue(-1).safemodeValue(0);
 CONFIG(bool, ShowLoadMessages).defaultValue(true);
-CONFIG(bool, LoadscreenLicenseLeftAlign).defaultValue(false);
 
 CLoadScreen* CLoadScreen::singleton = nullptr;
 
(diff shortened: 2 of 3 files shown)


ashdnazg:
It's bad to limit the notice to just 2 positions.

So:
1) This shouldn't be a config it should be a boolean you can set from luaintro.
2) When on, the notice won't be drawn by the engine.
3) The naming should be something along Spring.SetCustomLicenseNotice so it's obvious that the game's LuaIntro should draw that notice if the engine doesn't.
4) There will be consequences for a game removing the engine's notice without providing a suitable replacement (if you're unsure, ask).
5) Point 4 should be in the code's comments

Also, please don't refer to things regarding the license as crap. Many (including me) find this rather insulting, especially considering the past GPL-violations.

12345swordy:
May I ask what your motivation for this?

gajop:
Here's a political issue we can all discuss!
Personally I'm a a strong supporter of enforcing all licenses, and I will always side with adherence to Spring's GPL license.

However I agree with the idea behind this PR. It makes no sense to enforce printing this license. Why exactly *must* the players know how the engine is licensed, and certainly why should it be printed on the loadscreen? Not to mention that it can be wrong, as the game might have non-GPL licensed files (GPL-compat Lua and arbitrary license shaders). Then there's art...

The game definitely doesn't legally *need* to print out this information in any of its loading screens. These are not the terms of GPL (licenses are a set of rules, not a belief system).

abma:
IMHO the only real "benefit" of the license shown, that any game using the spring engine can easily be identified.

any person who wants to violate the license can easily remove this notice from the source code.

still, there are several other places where its clear which license the engine is:
- https://springrts.com/wiki/Download
- ~~README.markdown~~ ... sshh.. its missing there: mantised https://springrts.com/mantis/view.php?id=5211
- windows installer
- springlobby installer (indirectly as springlobby is gpl, too)
- each source code file in the first line
- a lot of other places in forum / wiki / other download pages

a person who redistributes the engine will have a problem to "not know" that its licensed as gpl.

so, the notice at start could be completely removed imo when 5211 is "fixed".

gajop:
Hey, so for completeness after some chat in sy I realized I might've been wrong in my previous statement about the license: http://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html (section 2.c)
It might require that it's printed visually in the game, at some point.
It's hard for me to interpret this (where should it be printed? must the engine control this? can it be moved from the original spot or left to the gamdevs? is Spring itself a derivative work of the "previous version of Spring" that had it added?), so I tried asking FSF for it. Here's the answer I got: http://hastebin.com/sotoseseke.xml , which is imo lacking any legal information and just ideological standpoint, so I'm not sure how to proceed on this...

abma:
show in info screen, too?

add to /help ?!

wrt pull-request, it should be LOG( ... not LOG_L(L_ERROR,

sprunk:
L_ERROR guarantees that the line is always there regardless of filtering level.

abma:
L_ERROR goes to stderr.
https://github.com/spring/spring/blob/develop/rts/System/Log/ConsoleSink.cpp#L43

there are a lot of other reasons to not use L_ERROR too, no discussion please, LOG() is correct / fits best.

ashdnazg:
I wish to remind that there still isn't consensus about this PR (far from it)

abma:
> I wish to remind that there still isn't consensus about this PR (far from it)

yep as this moves some responsbility to the game / person who distributes the game. also the gpl notice should be (re-)added to the default loading screen.

gajop:
After some thinking, my opinion on this is that we should retain the current GPL license in the loading screen as it might be a license violation if it was outright removed.

That said, it should still be possible for the game to remove it if they set a certain config value, but it should be **clearly** noted that in that case it is their responsibility to represent it in an "acceptable" (what is that?) fashion to the user. They can't just hide it and be done with it. It might need to be present in other loading screens, menus, about sections or similar (this is where GPL isn't very clear and most of their examples have to do with CLI interactive programs, rather than GUI or most certainly games).

abma:
> After some thinking, my opinion on this is that we should retain the current GPL license in the loading screen as it might be a license violation if it was outright removed.

define "removed". When you create some custom loadscreen, why should the engine printed license notice stay? also keep in mind that its shown for reload, too.
the default engine loading screen must have it, yes.

imo spring/gpl must be noted in "about" dialogs in gui and on the download page of the game at least.

gajop:
> When you create some custom loadscreen, why should the engine printed license notice stay
Odd as it may be, I think it should still stay even in those cases until you explicitly invoke a callout. A single Spring.SetDrawLicense(bool) would be sufficient. 

We don't want people to accidentally infringe on the GPL. I can totally imagine people that would want to just provide a visually interesting loadscreen but end up breaking the license because they didn't put any thought into it (and wouldn't have guessed the effect).

abma:
> We don't want people to accidentally infringe on the GPL. I can totally imagine people that would want to just provide a visually interesting loadscreen but end up breaking the license because they didn't put any thought into it (and wouldn't have guessed the effect).

" accidentally infringe on the GPL"... how should that happen?

they still must replace the default loading screen, so a "i didn't know that its gpl" is imo impossible. when downloading / installing spring from springrts.com you have to agree / easily can see that its licensed as gpl.

gajop:
If I understood your proposal correctly, you're suggesting that the license information should disappear as soon as people create a custom loadscreen (luaintro).

I think it's reasonable to think that just creating that isn't going to have any license implications. I believe that those two things 1) creating a loadscreen and 2) removing engine licensing should be two separate actions. We don't want people that do 1) to also immediately do 2). If they want to do both, they should do it explicitly (e.g. by invoking a callout). 

There are two additional things:
- all existing custom loadscreens would immediately cause the game to fail the 2.c clause of GPL.
- not everyone reads/knows the full text of GPL (and certainly not such obscure cases like this), and any things that might break licensing accidentally should be very clear, so it doesn't happen.

abma:
my intention is this:
spring is an engine: imo it should be possible and allowed to modify the load-screen at 100% without violating the license.

yes, a note must be printed that it is gpl and we already have serveral places where its shown (directly or indirectly) that it is gpl. the loadscreen is just an additional / prominet place, but it also lacks, that there is no easy way to read further info about it.

also the loading screen has the drawback that it can't be paused, so when spring loads to fast, the user sees nothing of it.

also 2c says "If the modified program normally reads commands interactively when run...": in the loading stage thats not possible! its possible when the game is loaded and enter is pressed.

gajop:
I agree with you.

> also the loading screen has the drawback that it can't be paused, so when spring loads to fast, the user sees nothing of it.

Regarding this, someone in the FSF suggested that the loadscreen should be 10s or so, allowing the user plenty of time to read the license info. I would be vehemently opposed to forcing any such changes as it would make development for Spring a nightmare. I'm sure most of you agree with me.

PS: Being able to control the loadscreen ("pause" it) can be good for some other reasons, such as getting desired the graphical effects.

sprunk:
2c also says

>(Exception: if the Program itself is interactive but does not normally print such an announcement, your work based on the Program is not required to print an announcement.)

-> if engine doesn't print anything in the loadscreen, games don't have to either and everything is fine.

gajop:
Well not really. The change to the engine itself would be a GPL infringement as we aren't the owners (no one really has full ownership of it, as there have been a number of contributors over the years)

Anarchid:
> Well not really. The change to the engine itself would be a GPL infringement as we aren't the owners (no one really has full ownership of it, as there have been a number of contributors over the years)

Spring should incorporate or subsume itself into SPI or SFC.

sprunk:
sigh

sprunk:
why reopen? GPL prevents the text from being removed

abma:
the text allows a lot of different interpretations / we didn't vote on this.

ashdnazg:
As far as I know jK's position on this is that a message must exist in a form that can't be hidden (hence the shadow under the white letters) and he's willing to allow a game to choose a suitable position out of multiple options.

My position is essentially this: https://github.com/spring/spring/pull/254#issuecomment-211914016 (allowing games to hide the default message and make it clear that they must put the same message themselves in a clear manner) but I believe that in issues such as license we should go with the strictest opinion in the team which is jKs (no option for hiding the message).
As to multiple options, I don't like that because it's avoiding the main issue that games have limited control over the loadscreen and will lead to clutter as every other game who wants a new place/font size/whatever will make a PR I have to check and merge.

**tl;dr: jK is against full customisation, I'm against partial customisation.**

Other options that will probably be accepted if someone implements them:
 * Temporary notice - fade it out after N seconds.
 * A "SpringPowered" splash screen.

In any solution it's probably possible to only show the notice on the first loading, so in a single-process-Chobby and the like you will only see it when the engine first starts.

abma:
1. shouldn't an engine allow full customisation?
2. there is no way to pause the loadscreen: so basicly the license info there is IMHO useless.
IMHO better:
- add Spring.GetLicense() which returns the whole license text so game devs can add it to the "about" dialog or into their splash screen
- log (the important parts) of the license to infolog.txt
- encourage game devs / content distributors to add a license info to their download pages

abma:
would be nice if @jk3064, @rtri, @hoijui could directly comment here.

sprunk:
My initial convo with jK:

>#moddev 2016-04-19 07:40 UTC

>Sprung also can i move the "spring 101.0, licensed under gpl blablabla" somewhere else?
>Sprung (or ideally, get rid of it completely)
>[LCC]jK Spring.GetConfigString("name")
>[LCC]jK and no spring brandmark is there intentionally
>[LCC]jK only possible replacement would be a short intro before LuaIntro
>[LCC]jK like nvidia does
>Sprung what about just changing alignment from center to left or right?
>[LCC]jK u are free to supply a patch that allows minimal customization without opening doors that hide the msg
>Sprung being on center means the corners are wasted because nothing really fits there, if the text was in one corner i could at least fit something into the other corner
>Sprung ok
>[LCC]jK i.e. the msg extra uses outline, so that you cannot put wite background behind it to make it invisible etc.

I dislike nvidia style intro because that still causes something else than my exact loadscreen to be shown, except worse: it now takes 100% of the screen instead of 10%.

gajop:
I don't think showing it just once for Chobby in the loadscreen would be good enough, as it loads so fast that I had to use screenshot capture to even see the new loadscreen.
I don't care so much what happens to the GPL notice, although I think I could make it look better if I could customize it.
The problem.. however is that if we could customize it and show it differently, I'm sure we'd have new discussions popup if someone were to make it a smaller/less visible font and thus essentially hide it. I think I'd rather avoid this and work on something more productive.

PR #256, opened 2016-05-05
Fix compile sqrt

LinLo:
Fix:
error : ‘sqrt’ is not a member of ‘std’
    const float r = std::sqrt((dx * dx) + (dy * dy)) / 64.0f;


abma:
seems cleanrock missed your pull requests, sorry. fixed in:

https://github.com/spring/spring/commit/238a9bfae28e6989f51b66e7ee6b7f2ad5be4aba

PR #257, opened 2016-05-11
Fix CUtils Project

12345swordy:


ashdnazg:
spring compiles great on VS 2015 at the moment.
I can't see how this is needed

12345swordy:
It's just a compile fix for one of the projects list that is generated by cmake. If you don't see a need for CUtils for msvc then modify the cmake file to remove it and close this pull request then.

PR #258, opened 2016-05-15
Increment path estimator version.

GoogleFrog:


GoogleFrog:
https://springrts.com/mantis/view.php?id=5234

PR #259, opened 2016-05-15
Added UnitDef's optional buildingMask tag

lhog:
Added UnitDef's optional buildingMask tag, essentially a 16 bit value.

By setting mask value in UnitDefs and on map (using new Spring.SetSquareBuildingMask synced control function) a mod/map author can mark down which structures (via UnitDef) could be placed on which squares of map (set in 2*SQUARE_SIZE coordinate space) and effectively block any other structure with non-matching mask.

By default both UnitDef's mask and map squares mask are set  to 1, so everything is buildable everywhere (other things like terrain steepness, water, enemy units, other buildings. etc. can still block construction as before)

Added as better alternative to https://github.com/spring/spring/pull/251
Use cases can be seen [there](https://github.com/spring/spring/pull/251) as well


ashdnazg:
Indeed a far better solution than the previous PR. I'll post some feedback later and this could probably be merged right after release

rts/Sim/Misc/BuildingMaskMap.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+	maskMap[x + z * mapDims.mapx] = value;
+	return true;
+}
+
+// tests previously set mask for tile[x,z] in SQUARE_SIZE coordinates against supplied value
+// true - construction is allowed, false - it's not
+bool BuildingMaskMap::TestTileMaskUnsafe(int x, int z, boost::uint16_t value)
+{
+	return value == maskMap[x + z * mapDims.mapx];
```
ashdnazg: since it's a mask, you probably want to do
```
const boost::uint16_t squareMask = maskMap[x + z * mapDims.mapx];
return value & squareMask == squareMask;
```

rts/Sim/Units/UnitDef.cpp
```diff
@@ -749,12 +752,33 @@ void UnitDef::CreateYardMap(std::string yardMapStr)
...
+	if (buildingMaskStart != std::string::npos) {
+		yardMapStrSize = buildingMaskStart + 1;
+
+		string numString = yardMapStr.substr(yardMapStrSize);
+		char* endPtr = nullptr;
+		errno = 0;
+		unsigned long number = strtoul(numString.c_str(), &endPtr, 0);
+
+		if (errno == ERANGE || number < 0 || number > USHRT_MAX) {			
+			LOG_L(L_WARNING, "%s: Given yardmap contains buildingMask size of greater than 16 bits \"%s\", setting buildingMask to default!", name.c_str(), numString.c_str());
```
ashdnazg: the errno and the dec/hex/bin is fun and all but unnecessary complication imo.

rts/Lua/LuaSyncedCtrl.cpp
```diff
@@ -3949,6 +3952,30 @@ int LuaSyncedCtrl::SetTerrainTypeData(lua_State* L)
 /******************************************************************************/
 /******************************************************************************/
 
+int LuaSyncedCtrl::SetBuildingMaskData(lua_State* L)
+{
```
ashdnazg: why not use luaL_checkint in the following?

rts/Lua/LuaSyncedCtrl.cpp
```diff
@@ -3949,6 +3952,30 @@ int LuaSyncedCtrl::SetTerrainTypeData(lua_State* L)
...
+	const int x = (int)luaL_checkfloat(L, 1);
+	const int z = (int)luaL_checkfloat(L, 2);
+	const float maskF = luaL_checkfloat(L, 3);
+	if (maskF < 0 || maskF > USHRT_MAX) {
+		luaL_error(L, "Incorrect value of mask: %s(%d, %d, %f)", __FUNCTION__, x, z, maskF);
+		lua_pushboolean(L, false);
+		return 0;
```
ashdnazg: you return 0 even though you push something to the stack...

rts/Lua/LuaSyncedCtrl.cpp
```diff
@@ -3949,6 +3952,30 @@ int LuaSyncedCtrl::SetTerrainTypeData(lua_State* L)
...
+	}
+	const boost::uint16_t mask = (boost::uint16_t)maskF;
+	const bool result = buildingMaskMap->SetTileMask(x, z, mask);
+	if (!result) {
+		luaL_error(L, "Invalid values supplied: %s(%d, %d, %d)", __FUNCTION__, x, z, mask);
+		lua_pushboolean(L, false);
+		return 0;
```
ashdnazg: same here

rts/Sim/CMakeLists.txt
```diff
@@ -16,7 +17,7 @@ add_library(engineSim STATIC
 		"${CMAKE_CURRENT_SOURCE_DIR}/Misc/DefinitionTag.cpp"
 		"${CMAKE_CURRENT_SOURCE_DIR}/Misc/GeometricObjects.cpp"
 		"${CMAKE_CURRENT_SOURCE_DIR}/Misc/GlobalSynced.cpp"
-		"${CMAKE_CURRENT_SOURCE_DIR}/Misc/GroundBlockingObjectMap.cpp"
+		"${CMAKE_CURRENT_SOURCE_DIR}/Misc/GroundBlockingObjectMap.cpp"
```
ashdnazg: whitespace

rts/Sim/Units/UnitDef.cpp
```diff
@@ -599,7 +599,8 @@ UnitDef::UnitDef(const LuaTable& udTable, const std::string& unitName, int id)
 	zsize = std::max(1 * SPRING_FOOTPRINT_SCALE, (udTable.GetInt("footprintZ", 1) * SPRING_FOOTPRINT_SCALE));
 
 	if (IsImmobileUnit()) {
-		CreateYardMap(udTable.GetString("yardMap", ""));
+		CreateYardMap(udTable.GetString("yardMap", ""));
```
ashdnazg: whitespace

jk3064:
I dislike the syntax, it breaks the yardmaps layout. Rule is 1 char = 1 square 
now you added '#...', so everytime you ask yourself where that square ends and next starts ....

ashdnazg:
in one of the commits he changed it to use a new buildingMask tag.
Should've edited the description too.

lhog:
Updated description.

PR #260, opened 2016-05-16
Update streflop_cond file to remove unneeded limitations

12345swordy:
Modify streflop/streflop_cond.h file as MSVC now has the std::isnan std::isinf and std::isfinite as part of the cmath header file.


ashdnazg:
1) It checks VS version, so it doesn't limit anything.
2) Don't touch streflop_cond.h or anything in streflop unless there's an extremely good reason
3) if you want to discuss this => find me at #sy

12345swordy:
I at sy right now, where are you?

PR #261, opened 2016-05-31
[AI] Add missing radarMiplevel and fix some docstrings

rlcevg:
According to [ModInfo.cpp](https://github.com/spring/spring/blob/6140fdde868985dbf05ba7d5126e17db5577664c/rts/Sim/Misc/ModInfo.cpp#L244) radarMipLevel is not constant (docstring says it's 8).


ashdnazg:
thank you!

PR #264, opened 2016-07-09
Add allowTake modrule (fix #5259)

rlcevg:
https://springrts.com/mantis/view.php?id=5259
related:
https://springrts.com/mantis/view.php?id=5257


ashdnazg:
thanks!

PR #265, opened 2016-07-18
Implement Spring.OpenURL

gajop:
Opening a discussion on Spring.OpenURL, this is related to https://github.com/Spring-Chobby/Chobby/issues/89

The use cases for this are mostly for the ingame lobby, with some examples:
- Opening a news/replays/forum, etc. link, that directs the user to the community.
- Clicking on a URL in text (this obviously could also be used for currently available ingame play).

Technically, this also makes it possible to open OS related URLs, such as directories or files, which isn't necessarily a bad thing imo.

So, thoughts? Any reasons against this, or should it be done differently?

### DIFF
--- a/rts/System/Platform/Misc.cpp
+++ b/rts/System/Platform/Misc.cpp
@@ -519,4 +519,20 @@ std::string ExecuteProcess(const std::string& file, std::vector<std::string> arg
 	return execError;
 }
 
+void OpenURL(const std::string& url)
+{
+#ifdef WIN32
+	ShellExecute(NULL, "open", url.c_str(), NULL, NULL, SW_SHOWNORMAL);
+#elseif __APPLE__
+	system(("open " + url).c_str()); // uhh
+#else
+	int pid = fork();
+	if (pid == 0) {
+		execl("/usr/bin/xdg-open", "xdg-open", url.c_str(), (char *)0);
+		//TODO: maybe use execl to find xdg-open?
+		exit(1);
+	}
+#endif
+}
+
 } // namespace Platform
--- a/rts/Lua/LuaUnsyncedCtrl.cpp
+++ b/rts/Lua/LuaUnsyncedCtrl.cpp
@@ -223,6 +223,8 @@ bool LuaUnsyncedCtrl::PushEntries(lua_State* L)
 	REGISTER_LUA_CFUNC(Restart);
 	REGISTER_LUA_CFUNC(Start);
 
+	REGISTER_LUA_CFUNC(OpenURL);
+
 	REGISTER_LUA_CFUNC(SetWMIcon);
 	REGISTER_LUA_CFUNC(SetWMCaption);
 
@@ -2202,6 +2204,13 @@ int LuaUnsyncedCtrl::Start(lua_State* L)
 		return 1;
 	}
 
+	return 0;
+}
+
+
+int LuaUnsyncedCtrl::OpenURL(lua_State* L)
+{
+	Platform::OpenURL(luaL_checkstring(L, 1));
 	return 0;
 }
 
--- a/rts/Lua/LuaUnsyncedCtrl.h
+++ b/rts/Lua/LuaUnsyncedCtrl.h
@@ -99,6 +99,8 @@ class LuaUnsyncedCtrl {
 		static int Restart(lua_State* L);
 		static int Start(lua_State* L);
 
+		static int OpenURL(lua_State* L);
+
 		static int SetWMIcon(lua_State* L);
 		static int SetWMCaption(lua_State* L);
 
--- a/rts/System/Platform/Misc.h
+++ b/rts/System/Platform/Misc.h
@@ -93,6 +93,12 @@ bool IsRunningInGDB();
  * @return error message, or "" on success
  */
 std::string ExecuteProcess(const std::string& file, std::vector<std::string> args, bool asSubprocess = false);
+
+/**
+ * Opens a URL in the default system application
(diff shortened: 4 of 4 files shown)


Anarchid:
- Games can (already, via widget handler poisoning) and should implement URL whitelisting.
- Perhaps there should be a springsettings entry (default false) for this. This way people can opt in (or opt out) and games can decide to ship the setting or not. They'd need to write-protect their springsettings from a widget-based attacker but that's another story.

rts/System/Platform/Misc.cpp
```diff
@@ -519,4 +519,20 @@ std::string ExecuteProcess(const std::string& file, std::vector<std::string> arg
...
+#ifdef WIN32
+	ShellExecute(NULL, "open", url.c_str(), NULL, NULL, SW_SHOWNORMAL);
+#elseif __APPLE__
+	system(("open " + url).c_str()); // uhh
+#else
+	int pid = fork();
+	if (pid == 0) {
+		execl("/usr/bin/xdg-open", "xdg-open", url.c_str(), (char *)0);
+		//TODO: maybe use execl to find xdg-open?
+		exit(1);
```
abma: doesn't indicate exit(1) an error which it isn't here?

rts/System/Platform/Misc.cpp
```diff
@@ -519,4 +519,20 @@ std::string ExecuteProcess(const std::string& file, std::vector<std::string> arg
 	return execError;
 }
 
+void OpenURL(const std::string& url)
```
abma: url should be parsed for beeing a valid url:

i.e. that things like rm -rf / aren't possible.

ashdnazg:
can't say I'm fond of this at all.

rts/System/Platform/Misc.cpp
gajop: I strongly dislike implementing non-trivial security checks myself, as it's easy to forget edge cases and those edge cases alone are enough to make the entire check pointless.
At the same time we'd lose some power, like being able to open directories, and I think these two operations are completely legitimate: 
`Spring.OpenURL(screenshotDir)`
`Spring.OpenURL(dataDir)`

Can/should we use a library to securely open URLs? 

Another alternative I'd like to consider is simply making games responsible for the things they do. If we restrict this to gadget unsynced, I think we can expect it to be used only by responsible people.

rts/System/Platform/Misc.cpp
gajop: Thanks! Much worse is the OSX part though. Haven't gotten around to implementing that yet.

gajop:
@ashdnazg can you be more specific please?

gajop:
@Anarchid using whitelists would perhaps defeat the purpose, as you couldn't open chat URLs.

Anarchid:
@gajop: my suggestion was whitelist patterns. In which nothing prevents one from either whitelisting every url with a http schema, or just `*`.

GoogleFrog:
I like the solution of allowing this for unsynced only. This creates no new player-originated exploits unless the game developer implements a lenient widget->unsynced gadget interface and enables local widgets. You've got to trust game developers to run their games.

gajop:
Perhaps the only problem with unsynced gadget approach is that maps could still contain such code, and we don't trust map makers in the same way (maps kinda count as player-distributed content)

Anarchid:
A map would be my preferred attack vector in all scenarios thus far. They are auto-downloaded, require no collusion from rapid admins, and can pretty much *overwrite* game data.

I'm not even sure if maps are not allowed to replace a game's widget handler.

jk3064:
Just tested `xdg-open file://bin/echo` it gives a warning and doesn't start binaries, so the worst case is already secured.

Still a widget could save/create a corrupted JPG that cause a buffer overrun or something and hack the computer that way.

So I would still prefer to disable `file://` links, even when I would see a few useful use-cases for it.

Not to forget that windows' `open` is very likely not as secure as xdg-open is, and making windows' safe is likely a huge work.

Anarchid:
If malicious jpg are a threat, what's to prevent me from including a corrupted jpg as as a texture of a map's included feature, completely bypassing Spring Lua and attacking DevIL directly?

abma:
> If malicious jpg are a threat, what's to prevent me from including a corrupted jpg as a texture of a map's included feature, completely bypassing Spring Lua and attacking DevIL directly?

basicly nothing. imo that was just a bad example.

When the function OpenURL is implemented we should take care that only an url in a browser can be opened with the function and nothing else.

Anarchid:
Limitting OpenUrl to http and https schemas would close most evil cases not already handled by OS itself, i guess.

I'll check if Mac allows launching executables via `open` tomorrow; my suspicion is that it's at least as strict as `xdg-open`.

Anarchid:
Turns out naive `open` on Mac is utterly terrifying: `open ~/.spring/engine/102.0/Spring_102.0.app/Contents/MacOS/spring` launched spring without confirmation or prompt.

rts/System/Platform/Misc.cpp
tomjn: A regular expression check of the string should verify if it is indeed a valid URL. If we're being strictly OO, the parameter should be a dedicated URL object not an `std::string`

tomjn:
My concern would be less with maps and games, but malicious links inside lobby channels

gajop:
There is unlikely to be a satisfying engine implementation from what I can tell, and interested parties have moved towards using wrappers that expose similar functionality. Closing.

PR #268, opened 2016-09-02
Fix wrong maphelper default splat values

sprunk:
See https://github.com/spring/spring/blob/develop/rts/Map/MapInfo.cpp#L168


ashdnazg:
thanks!

PR #269, opened 2016-09-13
Add option to force thread sleep (for chobby idle cpu usage)

Licho1:
+ fix unitsync export defs + windows compatilation issue


rts/Menu/LuaMenuController.cpp
```diff
@@ -80,6 +79,9 @@ bool CLuaMenuController::Draw()
 	// calls IsAbove
 	mouse->GetCurrentTooltip();
 
+	
+	spring_msecs(10).sleep(true);
```
ashdnazg: why did you move the location of the sleep?
(I'm quite certain it should be where it was)

Licho1: Imo it should be near the step that can take lots of time (drawscreen). But it does not really matter, I was just moving it around to test some things with even bigger delay. Move it wherever you prefer

ashdnazg:
Merged in 5378bbb7dc539403275d0f442d0a1cb1fd7dfd58 and 83ea02727ffa8c67daaf32515c771f9490304cbc

Thanks!

PR #270, opened 2016-09-13
Unhardcode leash distances

sprunk:


ashdnazg:
maneuverLeash wasn't used in anything?

sprunk:
It was only used here: https://github.com/spring/spring/blob/develop/rts/Sim/Units/CommandAI/MobileCAI.cpp#L665

ashdnazg:
did you test how units behave afterwards?

sprunk:
Not yet

GoogleFrog:
I am worried that you have replaced many different magic numbers with one configurable value.  For example MobileCAI uses three different magic numbers and the difference between these numbers may have been important. A change like this should always implement the old behavior by default so I recommend using a new unitDef (because games will have deprecated tags hanging around in them) and preserving the ratio of  the magic numbers.

PR #273, opened 2016-10-02
Update JVM Memory limits to current decade

DeinFreund:
64MB runs GC every frame..

### DIFF
--- a/AI/Interfaces/Java/data/jvm.properties
+++ b/AI/Interfaces/Java/data/jvm.properties
@@ -68,8 +68,8 @@
 ;jvm.option.x=-Djava.library.path=...
 
 # footprint (memory) related
-jvm.option.x=-Xms16M
-jvm.option.x=-Xmx64M
+jvm.option.x=-Xms64M
+jvm.option.x=-Xmx512M
 jvm.option.x=-Xss512K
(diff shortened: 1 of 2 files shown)


abma:
thanks! (i tested squashing the commits via github)

for the curious: did/do you ever change jvm.useDebugOptions in https://github.com/spring/spring/blob/develop/AI/Interfaces/Java/data/jvm.properties#L32

IMHO this can be removed as renaming a file is basicly the same work as changing an option.

DeinFreund:
I never messed with this config at all before running into those GC issues. But that switch seems kinda pointless, as you said one could just rename the file.

Btw, is there a way to include custom properties just for a single AI? Or does it have to be changed for the whole interface.

abma:
> Btw, is there a way to include custom properties just for a single AI? Or does it have to be changed for the whole interface.

atm there is no way afaik. i would also guess that there is only a single instance of the jvm loaded:
https://github.com/spring/spring/blob/develop/AI/Interfaces/Java/src/main/native/JavaBridge.c#L68

Not sure if its possible at all to load multiple jvm instances in a single process as jvm i.e. changes a lot of "global" things in a process (i.e. it changes signal handlers).

PR #276, opened 2016-10-15
add callin AllowDraw for LuaMenu

Licho1:
* Adds AllowDraw for LuaMenu to allow FPS rate limiting from lua (for idling in chobby)
 * Stop rendering when window is minimized


ashdnazg:
Looks good I think, I just want to verify this was tested (especially the SDL events stuff) with a game running as well

abma:
hasn't the config setting MinimizeOnFocusLoss=1 the same or at least similar effect?

ashdnazg:
I don't think it does.

ashdnazg:
Licho, your last commit duplicates work used in Update. If you want to do this correctly, check if drawing is needed in LuaMenuController::Update (call there all stuff that should happen whether drawing occurred or not) and use the value returned from there.
It's possible that the swapbuffer code can be passed into the 
```
if (ret) {
			ret = activeController->Draw();
}
```
block

Licho1:
Thats not the logic it uses though. If update returns fals, spring shuts down, it stops its main cycle. It was never used though but I didnt want to change that function.

Licho1:
It does sound cleaner though. Stuff could be separated cleanly for update and draw and update deciding whether draw should happen.

ashdnazg:
hmm, you are correct.
It may be nice if you could change https://github.com/spring/spring/blob/develop/rts/Game/Game.cpp#L1050
to 'return false' as well, and see if it works

ashdnazg:
Thanks Licho!

PR #277, opened 2016-10-30
Update sent and received values on NETMSG_AISHARE

rlcevg:
Allows receiver-AI to better operate with its income (based on resources going from sender-AIs).
Makes NETMSG_AISHARE seem more like NETMSG_SHARE.


ashdnazg:
very nice catch!

PR #279, opened 2016-11-25
Maximum height of overhead camera is now configurable.

GoogleFrog:


GoogleFrog:
Tested and works.

cleanrock:
I have thought about adding this config too, it works well but i wonder if we should use this factor in OverviewController too, if so, this change seems ok to me:

```
diff --git a/rts/Game/Camera/OverviewController.cpp b/rts/Game/Camera/OverviewController.cpp
index 4576b92..c3f325a 100644
--- a/rts/Game/Camera/OverviewController.cpp
+++ b/rts/Game/Camera/OverviewController.cpp
@@ -9,6 +9,7 @@
 #include "Game/UI/MouseHandler.h"
 #include "Sim/Misc/GlobalConstants.h"
 #include "System/Log/ILog.h"
+#include "System/Config/ConfigHandler.h"
 
 COverviewController::COverviewController()
 {
@@ -17,7 +18,7 @@ COverviewController::COverviewController()
 
        pos.x = mapDims.mapx * 0.5f * SQUARE_SIZE;
        pos.z = mapDims.mapy * 0.5f * SQUARE_SIZE;
-       const float height = std::max(pos.x / globalRendering->aspectRatio, pos.z);
+       const float height = std::max(pos.x / globalRendering->aspectRatio, pos.z) * configHandler->GetFloat("OverheadMaxHeightFactor");
        pos.y = CGround::GetHeightAboveWater(pos.x, pos.z, false) + (2.5f * height);
 
        dir = float3(0.0f, -1.0f, -0.001f).ANormalize();

```

ashdnazg:
If you're working on camera settings, can you please make them changeable on the fly?
Check the FPS cam for an example.

GoogleFrog:
@sprunk can you look into this since you seem to be able to compile Spring?

sprunk:
Sure, can try

sprunk:
#293

ashdnazg:
> @sprunk can you look into this since you seem to be able to compile Spring?

This is funny because @sprunk doesn't compile spring but edits the files on github.

PR #282, opened 2016-12-19
Expose weapon salvo vars to Lua SetUnitWeaponState

sprunk:


ashdnazg:
I can't check at the moment, but if the vars don't appear in the equivalent Get function, please add them there as well.

ashdnazg:
Thanks!

PR #288, opened 2017-01-31
Remove old warning that no longer appears

sprunk:
Halley's Comment - next appearance on Spring 140!


rtri:
~four years is probably long enough of a grace period, ta

PR #289, opened 2017-02-07
fix indentation

sprunk:


rtri:
thanks, but fixed in 3f7561d7be87ca132e9ce0b8482ca30bd72d4160.

PR #290, opened 2017-02-07
AllowStartPosition now receives 9th parameter teamID

sprunk:
Useful for AIs who always report playerID 255 (also convenience)


rtri:
I think it's better to just bite the bullet and sanitize `gadget:AllowStartPosition`'s parameter order.

rtri:
That's fine.

sprunk:
So what should the order be?
`playerID, teamID, readyState, real xyz, clamped xyz` ?

rtri:
The same as the engine's internal order: player, team, ready, clamped, raw.

rtri:
looks good, thanks.

PR #292, opened 2017-02-17
Don't count specs when deciding the "correct" sync

sprunk:


ashdnazg:
1) the correct sync doesn't mean anything anyway
2) spectators are far less likely to cheat and desync on purpose, so they're more reliable than players for that matter.

My instinct is no-merge

rtri:
Hmm, not a fan. It's useful to factor in the silent majority, for example in a 1v1 with 8 specs.

sprunk:
What if the two 1v1 players are in sync with each other? Their game is valid. How about specs do count if there is desync between players, but are always considered the minority if all players are in sync between one another?

ashdnazg:
If you want, you can go through the demo and check the checksums to see who desynced with whom.
That's the correct way to determine what you want.
The Logs don't work this way, you may get sync error for both players first, then the specs will have errors and then the players will have them back. Unreliable - hence no meaning for "correct".

sprunk:
Mostly I just wanted to prevent the warning if it doesn't affect real players (so as not to cause chaos in running games, replay data won't help here).

This can also be done gameside though so no biggie.

ashdnazg:
last thing I want to do is to prevent the warning!

sprunk:
The specs would still be considered desynced and receive the message. Just that specs get the message as a PM and it's not shown to players.

ashdnazg:
Still, what you plan won't work, and will have cases where spec desyncs are reported as players desync and vice versa.
Also we're going to have people not report desyncs because "it was solved" or something, because the message stopped appearing.
That's very bad.
As a rule, hiding warnings is a terrible idea.

PR #293, opened 2017-02-17
Overhead camera config:

sprunk:
* max height now configurable (closes #279)
 * changeable on the fly

### DIFF
--- a/doc/changelog.txt
+++ b/doc/changelog.txt
@@ -35,6 +35,7 @@ Lua:
 UI:
  - Add FPSClampPos config for whether the engine should verify the camera doesn't hit ground/go too far
    this and other FPS camera configs can be changed midgame and will affect the next frame drawn.
+ - Add OverheadMaxHeightFactor config - a float multiplier for maximum overhead camera height.
 
 -- 103.0 --------------------------------------------------------
 General:
--- a/rts/Game/Camera/OverheadController.cpp
+++ b/rts/Game/Camera/OverheadController.cpp
@@ -20,7 +20,7 @@ CONFIG(int, OverheadScrollSpeed).defaultValue(10);
 CONFIG(float, OverheadTiltSpeed).defaultValue(1.0f);
 CONFIG(bool, OverheadEnabled).defaultValue(true).headlessValue(false);
 CONFIG(float, OverheadFOV).defaultValue(45.0f);
-
+CONFIG(float, OverheadMaxHeightFactor).defaultValue(1.0f);
 
 static const float angleStep = math::HALFPI / 14.0f;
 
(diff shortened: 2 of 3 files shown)


GoogleFrog:
Is there anything blocking this? Low camera height is getting frustrating.

rtri:
In general, there are far too many config keys for the default cameras already.

A proper `LuaCameraController` seems like a better idea: what if someone suddenly starts caring about an `OverheadMinHeightFactor`, or an `OverheadMaxFOV`, or ...?

GoogleFrog:
If they care about those things they can make PRs about them. Is Spring supposed to implement a useful default camera or not?

rtri:
Such PR's would be rejected, and kindly don't resort to disingenuous arguments by implying the default is somehow useless for lacking one single bit of polish.

GoogleFrog:
That isn't quite what I mean by useful. I mean to ask whether polished games are meant to use a Spring-implemented camera or whether such games are expected to implement their own with lua.

When a gamedev wants something slightly more than is offered by the engine they have two options:
 * Try to get a PR pulled for an extension of the relevant feature.
 * Reimplement the entire feature in lua.

This is a common occurrence. Reimplementing it in lua wastes a lot of time and often has limitations not present in the engine (I'm not impressed with any lua camera implementation). Engine implementations do incur some debt but help all games and tend to be more reliable.

A related issue is that the replacement of an engine feature with lua often leads to bitrot. This makes Spring less approachable as new developers either have to track down multiple replacement widgets or deal with broken defaults.

rtri:
If a game is not satisfied with Spring's default implementation of some feature and the available Lua interface(s) are powerful enough, then expecting a Lua replacement implementation is probably fair. This has already happened several times anyway, e.g. with the UI.

If an interface is lacking (this may be the case for cameras) and the default is almost-but-not-quite-satisfactorily implemented according to enough games making use of it, then my position would be to address the API first because that benefits many more projects in the long-term and often also allows further engine cleanup (countering bitrot) once a solid Lua solution becomes widespread.

Usually reimplementing a feature in Lua has fewer limitations, so what specifically makes cameras troublesome?

ashdnazg:
@rtri while I 100% agree with what you say, I kinda implicitly agreed to merge this here: https://github.com/spring/spring/pull/279#issuecomment-263085504 and in chat.
It wouldn't be right to reject this after making them update the PR (even though the update was rather trivial.

doc/changelog.txt
```diff
@@ -35,6 +35,7 @@ Lua:
 UI:
  - Add FPSClampPos config for whether the engine should verify the camera doesn't hit ground/go too far
    this and other FPS camera configs can be changed midgame and will affect the next frame drawn.
+ - Add OverheadMaxHeightFactor config - a float multiplier for maximum overhead camera height.
```
abma: FYI: for documentation of config values  .description() exists. i.e.

https://springrts.com/wiki/Springsettings.cfg#3DTrees
https://github.com/spring/spring/blob/403f6cb157da37bbb820611bae40b09f544f1126/rts/Rendering/Env/ITreeDrawer.cpp#L26

PR #294, opened 2017-02-19
Circling construction plane fixes

CommanderSpice:
Aircons in Phoenix Annihilation (see https://github.com/CommanderSpice/Phoenix-Annihilation) have collision disabled.
They don't do proper circling without the modification in BuilderCAI.cpp. Instead they just clump together above the unit beeing build.

The other Modification randomizes the circling direction, so they kind of distribute.
The distribution looks like a circle when collision is disabled. With collision they use a spherical formation.


rtri:
thanks for restoring this behavior, was broken during a refactor years ago.

PR #295, opened 2017-02-25
[AI] Fix direct mem leak @ResourceHandler.cpp:137

rlcevg:
[One of the mem-leaks reported by AddressSanitizer](https://github.com/spring/spring/blob/57759758537a8110b5b8f0c1e2500ab0308ba985/rts/Sim/Misc/ResourceHandler.cpp#L137)


rtri:
ty

PR #296, opened 2017-02-26
Fix floater tag assignment

rlcevg:
Despite having "floater=true" in [luaDef](https://github.com/ZeroK-RTS/Zero-K/blob/master/units/shipaa.lua#L35) AI API UnitDef::IsFloater() returns false, which means floatOnWater was not properly set.
[old behaviour](https://github.com/spring/spring/blob/af1edb128b06fd32c50a767f48a7ca6b58450ae2/rts/Sim/Units/UnitDef.cpp#L534)
[new broken behaviour](https://github.com/spring/spring/blob/develop/rts/Sim/Units/UnitDef.cpp#L533)
I didn't find a trace of MoveCtrl.SetMoveDef in ZK.

I don't know how to properly fix it, but this PR works for me.

### DIFF
--- a/rts/Sim/Units/UnitDef.cpp
+++ b/rts/Sim/Units/UnitDef.cpp
@@ -523,7 +523,6 @@ UnitDef::UnitDef(const LuaTable& udTable, const std::string& unitName, int id)
 
 		if (moveDef == nullptr) {
 			upright           |= !canfly;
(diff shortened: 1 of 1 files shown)


rtri:
`floatOnWater` is only configurable / parsed for aircraft and buildings.

AI's should check the MoveDef `speedModClass` property.

PR #297, opened 2017-02-26
[AI] Fix UnitDef::IsFloater()

rlcevg:
Return proper floatOnWater.

~~RAI's (legacy API) naval ability will remain broken.~~
Damn those AI API layers. RAI also fixed.


rtri:
Thanks.

Future maintainability wasn't a high priority during the AI interface rewrite, and these issues are part of the price we have to pay for it.

PR #298, opened 2017-03-01
Fix a constant - nlc

sprunk:
Was supposed to be 20 degrees but is 20 radians.
20 radians is 3*360 + 20 degrees (!) so logic doesn't change but this prevents failures if someone changes the value.

Relatedly, this whole check might be pointless because there's already the weaponDef->maxFireAngle check and that tag also defaults to 20 degrees.

### DIFF
--- a/rts/Sim/Weapons/Weapon.cpp
+++ b/rts/Sim/Weapons/Weapon.cpp
@@ -343,7 +343,7 @@ void CWeapon::ReAimWeapon()
 
 	// check max FireAngle
 	reAim |= (wantedDir.dot(lastRequestedDir) <= weaponDef->maxFireAngle);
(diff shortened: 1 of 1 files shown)


ashdnazg:
you seem to misunderstand what nlc means

ashdnazg:
amusing!

ashdnazg:
btw, 20 radians appears to be 65 degrees

gajop:
these must be radians in the imperial system

sprunk:
Oops, that's right. Missed a pi somewhere.

rtri:
20 / 2pi ~= 3.18 circumferences = 0.18*2pi radians ~= 66 degrees.

PR #300, opened 2017-03-12
Added DrawScreenFinal callin

lhog:
Added DrawScreenFinal(vsx, vsy) unsynced callin. It's used to alter content of the frame after it's been completely rendered (i.e. World, MiniMap, Menu, UI).
Basically the frame you would get if you /screenshot it. Use it similarly to DrawScreenEffects (by ivand).

The specific use case I had in mind would be screen space color correction for color blind persons.

The alternative (existing) approach is complicated, if doable at all. While the "World" part of the screen can easily be modified by DrawScreenEffects()
* Most of the UI usually drawn via Chili right after the DrawScreenEffects() has been called and Chili doesn't seem to have an easy way to apply shaders to it's components.
* A lot of small UI components are drawn via old-school gl.Line, gl.Rect kind of calls, which also makes impractical handling such cases individually.


rtri:
Spring's convention is `Post`, not `Final`.

The duplication in `LuaOpenGL::*DrawScreen*` and `CLuaHandle::DrawScreen*` should also be removed.

lhog:
> Spring's naming convention for this type of callin is ```Post```, not ```Final```.

That's clear. Will do

> The duplication in ```LuaOpenGL::*DrawScreen*``` and ```CLuaHandle::DrawScreen*``` should also be removed.

Please elaborate the preferred way, if any: i can move similar pieces to separate functions or intorduce a preprocessor definition.

rtri:
In `LuaOpenGL`, add a `DrawScreenCommon` and make the other functions call that.

In `CLuaHandle`, factor out the common part of `DrawScreen` into a similar common function that takes a `LuaHashString` as its argument.

rtri:
good enough, thx.

PR #301, opened 2017-03-20
CEG: add 'underground' bit

sprunk:


rtri:
While this might have some use by itself on voidground maps, I'd simply change the definition of underground `(altitude  < -1.0f)` to be less strict.

PR #302, opened 2017-03-20
Lasercannon: allow direction change

sprunk:
The most important use case is allowing ground bounces (currently it keeps grinding against the ground even if set to bounce)

### DIFF
--- a/rts/Sim/Projectiles/WeaponProjectiles/LaserProjectile.cpp
+++ b/rts/Sim/Projectiles/WeaponProjectiles/LaserProjectile.cpp
@@ -42,7 +42,6 @@ CLaserProjectile::CLaserProjectile(const ProjectileParams& params): CWeaponProje
 {
 	projectileType = WEAPON_LASER_PROJECTILE;
 
-	// FIXME: constant, assumes |speed| never changes after creation
 	speedf = speed.w;
 
 	if (weaponDef != NULL) {
@@ -131,7 +130,7 @@ void CLaserProjectile::UpdatePos(const float4& oldSpeed) {
 	UpdateGroundBounce();
 
 	if (oldSpeed != speed) {
-		SetVelocityAndSpeed(speed);
+		speedf = speed.w;
 	}
 }
 


ashdnazg:
This patch seems very very wrong.
When speed is changed, SetVelocityAndSpeed should be called. That's a rule

sprunk:
That's already done by UpdateGroundBounce() here https://github.com/sprunk/spring/blob/94ce59dfb49cdbfa4bac73500b433eaee8ee0711/rts/Sim/Projectiles/WeaponProjectiles/WeaponProjectile.cpp#L317

ashdnazg:
but you've edited UpdatePos which is called by other places as well

sprunk:
It doesn't matter where it is called from because the speed change is encapsulated:

<Something, doesn't matter what> calls UpdatePos. UpdatePos calls UpdateGroundBounce.

UpdateGroundBounce changes 'speed', calls SetVelocityAndSpeed, and returns control to UpdatePos.

UpdatePos only updates some cached value 'speedf'.

ashdnazg:
but speedf was never updated before, why do you assume it's a cached value.
This won't get merged

sprunk:
After some thinking this patch does not do the thing it was supposed to (I'd argue it's still correct but no point anymore).

ashdnazg:
please invest the time in creating a compilation environment on your PC and test your patches.

rtri:
>> currently it keeps grinding against the ground even if set to bounce

LaserCannon bouncing works fine for me in local tests so this would either indicate a bug or tag misuse.

Anarchid:
Simulation works fine, yes. Visuals produced this last time i attempted:
image

sprunk:
In my test I took ZK corsh and changed its weapontype from Cannon to Lasercannon.
It has groundbounce, numbounce = 40, noExplode.

I increased peewee radius to 25 to make very sure it covers the hitvolume (shouldnt be needed), and disabled weapon_single_hit.lua gadget which does something unrelated with the weapon (it only modifies damage on hit but just in case).

screen00064

The bounce effect happens between the 1st and 2nd peewee but no peewee is hit. I thought maybe it goes under them (the weapon has only 0.15 vertical bounce) but the hitvolume goes all the way to the ground:
screen00063

This also happens with Cannon but feels much easier to reproduce with LaserCannon. Depending on the angle, sometimes a peewee behind is hit but not the ones in front of it.

PR #304, opened 2017-03-24
Added SetUnitWaterline LUA function

sanguinariojoe:
Right now, in order to create a submarine which may become surfaced, or even submerged at different levels, you must create several objects with different waterline parameters, morphing/replacing the unit each time. Hence, adding the LUA function SetUnitWaterline may become very convenient.

As far as I know, this patch should not affect the current games at all (since the waterline parameter is taken from the unit definition at PreInit() method)

### DIFF
--- a/rts/Sim/MoveTypes/GroundMoveType.h
+++ b/rts/Sim/MoveTypes/GroundMoveType.h
@@ -41,8 +41,6 @@ public:
 	bool CanApplyImpulse(const float3&) override;
 	void LeaveTransport() override;
 
-	float Waterline() const override { return waterline; }
-
 	void InitMemberData();
 	bool SetMemberValue(unsigned int memberHash, void* memberValue) override;
 
@@ -189,7 +187,6 @@ private:
 	bool canReverse;
 	bool useMainHeading;                /// if true, turn toward mainHeadingPos until weapons[0] can TryTarget() it
 	bool useRawMovement;                /// if true, move towards goal without invoking PFS
-	float waterline;
 
 	float skidRotSpeed;                 /// rotational speed when skidding (radians / (GAME_SPEED frames))
 	float skidRotAccel;                 /// rotational acceleration when skidding (radians / (GAME_SPEED frames^2))
--- a/rts/Sim/MoveTypes/GroundMoveType.h
+++ b/rts/Sim/MoveTypes/GroundMoveType.h
@@ -41,7 +41,6 @@ public:
 	bool CanApplyImpulse(const float3&) override;
 	void LeaveTransport() override;
 
-	bool FloatOnWater() const override { return floatOnWater; }
 	float Waterline() const override { return waterline; }
 
 	void InitMemberData();
@@ -145,7 +144,7 @@ private:
 	bool WantReverse(const float3& wpDir, const float3& ffDir) const;
 
 private:
-	std::array<std::pair<unsigned int,  bool*>, 3>  boolMemberData;
+	std::array<std::pair<unsigned int,  bool*>, 2>  boolMemberData;
 	std::array<std::pair<unsigned int, short*>, 1> shortMemberData;
 	std::array<std::pair<unsigned int, float*>, 9> floatMemberData;
 
@@ -190,8 +189,6 @@ private:
 	bool canReverse;
 	bool useMainHeading;                /// if true, turn toward mainHeadingPos until weapons[0] can TryTarget() it
 	bool useRawMovement;                /// if true, move towards goal without invoking PFS
-
-	bool floatOnWater;
 	float waterline;
 
 	float skidRotSpeed;                 /// rotational speed when skidding (radians / (GAME_SPEED frames))
--- a/rts/Sim/MoveTypes/GroundMoveType.h
+++ b/rts/Sim/MoveTypes/GroundMoveType.h
@@ -41,6 +41,8 @@ public:
 	bool CanApplyImpulse(const float3&) override;
 	void LeaveTransport() override;
 
+	bool FloatOnWater() const override { return floatOnWater; }
+
 	void InitMemberData();
 	bool SetMemberValue(unsigned int memberHash, void* memberValue) override;
 
@@ -142,7 +144,7 @@ private:
 	bool WantReverse(const float3& wpDir, const float3& ffDir) const;
 
 private:
-	std::array<std::pair<unsigned int,  bool*>, 2>  boolMemberData;
+	std::array<std::pair<unsigned int,  bool*>, 3>  boolMemberData;
 	std::array<std::pair<unsigned int, short*>, 1> shortMemberData;
 	std::array<std::pair<unsigned int, float*>, 8> floatMemberData;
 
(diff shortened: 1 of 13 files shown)


rtri:
Nothing against exposing it, but this would fit better within the `MoveCtrl.SetGroundMoveTypeData` API.

rts/Sim/Units/Unit.cpp
```diff
@@ -399,6 +401,8 @@ void CUnit::PreInit(const UnitLoadParams& params)
 		selfdExpDamages = DynDamageArray::IncRef(&unitDef->selfdExpWeaponDef->damages);
 	if (unitDef->deathExpWeaponDef != nullptr)
 		deathExpDamages = DynDamageArray::IncRef(&unitDef->deathExpWeaponDef->damages);
+
+    waterline = unitDef->waterline;
```
sprunk: whitespace

rts/Sim/Units/Unit.h
```diff
@@ -518,6 +518,9 @@ class CUnit : public CSolidObject
 	UnitTrackStruct* myTrack;
 	icon::CIconData* myIcon;
 
+    // Used to allow LUA to modify per-object waterline parameter
+    float waterline;
```
sprunk: whitespace

sanguinariojoe:
To be honest, my first impulse was to add it to `MoveCtrl.SetGroundMoveTypeData` function, but while developing I changed my mind... I guess I don't like the `MoveCtrl` API

Anyway you are right, I'll extend `MoveCtrl.SetGroundMoveTypeData` to support waterline (and eventually remove SetUnitWaterline)

rts/Sim/Units/Unit.cpp
sanguinariojoe: Fixed

rts/Sim/Units/Unit.h
sanguinariojoe: Fixed

sanguinariojoe:
Now I remember why I decided don't using `MoveCtrl.SetGroundMoveTypeData`...
First, `floatOnWater` has been removed from such function. I'm afraid the comment 
> NOTE: not all keys and not all values may work correctly.
reflects precisely this fact.
Second, waterline is used both by `GroundMoveType` and `StaticMoveType`. However `StaticMoveType` cannot be controlled by the methods provided in `MoveCtrl`
Hence, I guess that `SetUnitWaterline` and `GetUnitWaterline` (added in https://github.com/spring/spring/pull/304/commits/826566637a0abf3e1d4559ba3b5367185d304f9d) may be more convenient (and simpler)

rtri:
>> `floatOnWater` has been removed from such function

Not quite, `floatOnWater` has never been *settable* through these functions.

>> However `StaticMoveType` cannot be controlled by the methods provided in `MoveCtrl`

Not really important so long as `waterline` becomes a member of `AMoveType`, you can either add a `SetStaticMoveTypeData` callout or just ignore that for now.

sanguinariojoe:
OK, I implemented waterline in `MoveCtrl.SetGroundMoveTypeData`, as well as `floatOnWater`. I also removed `SetUnitWaterline` and `GetUnitWaterline`.

rts/Sim/Units/Unit.cpp
```diff
@@ -1666,6 +1662,15 @@ bool CUnit::FloatOnWater(bool onlyDef) const {
 	return (unitDef->floatOnWater);
 }
 
+float CUnit::Waterline() const {
+	// Check if a move type is already in charge
+	if (moveType != nullptr)
```
rtri: `moveType` is never null while a unit is alive, this won't do what you want.

rtri:
Please keep the `floatOnWater` changes separate from this PR.

`floatOnWater` wasn't configurable yet because doing so more or less forces rethinking the `MoveDef` concept, which uses it as a hard boundary between ships/hovers and tanks/bots. Those are also tied to different `MoveMath` code.

Some stylistic notes wrt `waterline` diffs to follow.

rts/Sim/MoveTypes/GroundMoveType.h
```diff
@@ -189,7 +190,9 @@ class CGroundMoveType : public AMoveType
 	bool canReverse;
 	bool useMainHeading;                /// if true, turn toward mainHeadingPos until weapons[0] can TryTarget() it
 	bool useRawMovement;                /// if true, move towards goal without invoking PFS
+
 	bool floatOnWater;
+	float waterline;
```
rtri: can be moved to `AMoveType` under `float maneuverLeash`.

rts/Sim/MoveTypes/GroundMoveType.h
```diff
@@ -42,6 +42,7 @@ class CGroundMoveType : public AMoveType
 	void LeaveTransport() override;
 
 	bool FloatOnWater() const override { return floatOnWater; }
+	float Waterline() const override { return waterline; }
```
rtri: can be moved to `AMoveType` under `float GetManeuverLeash() const`.

rts/Sim/MoveTypes/MoveType.cpp
```diff
@@ -129,3 +129,11 @@ bool AMoveType::SetMemberValue(unsigned int memberHash, void* memberValue) {
 	return false;
 }
 
+
+float AMoveType::Waterline() const {
+	// creg
+	if (owner == nullptr)
+		return 0.0f;
+
+	return (owner->unitDef->waterline);
```
rtri: 1) owner can never be `null` at the point where this is called.
2)  initialize `waterline` to `owner->unitDef->waterline` in `AMoveType`'s constructor and return it here.

rts/Sim/MoveTypes/StaticMoveType.cpp
```diff
@@ -10,12 +10,15 @@
...
+	CR_IGNORED(floatMemberData),
+	CR_MEMBER(floatOnWater),
+	CR_MEMBER(waterline)
 ))
 
 CStaticMoveType::CStaticMoveType(CUnit* owner):
 	AMoveType(owner),
-	floatOnWater(false)
+	floatOnWater(false),
+	waterline(0.0f)
```
rtri: same as above.

rtri:
just two points left now:

1) `MoveType.cpp` is missing the `CR_MEMBER(waterline)` declaration
2) the actual assignment to `waterline` in `CGroundMoveType::SetMemberValue` can be handled in `AMoveType::SetMemberValue`

rts/Sim/MoveTypes/MoveType.cpp
```diff
@@ -42,7 +42,8 @@ AMoveType::AMoveType(CUnit* owner):
 	maxSpeedDef(owner? owner->unitDef->speed / GAME_SPEED : 0.0f),
 	maxWantedSpeed(owner? owner->unitDef->speed / GAME_SPEED : 0.0f),
 
-	maneuverLeash(500.0f)
+	maneuverLeash(500.0f),
+	waterline(owner->unitDef->waterline)
```
rtri: should be `waterline(owner? owner->unitDef->waterline: 0.0f)` due to the workings of CREG

rts/Sim/MoveTypes/StaticMoveType.h
```diff
@@ -22,16 +22,11 @@ class CStaticMoveType : public AMoveType
 	bool Update() override { return false; }
 	void SlowUpdate() override;
 
-	float Waterline() const override { return waterline; }
-
 	void InitMemberData();
 	bool SetMemberValue(unsigned int memberHash, void* memberValue) override;
 
 private:
 	std::array<std::pair<unsigned int, float*>, 1> floatMemberData;
```
rtri: obsolete

sanguinariojoe:
OK! I also rebased the branch to can safely add a note at the changelog

rtri:
alright, thanks :)

PR #307, opened 2017-04-15
Allow paralyzers to damage features

sprunk:
Previously Feature[Pre]Damaged wasn't even called so no easy way to emulate.

### DIFF
--- a/rts/Sim/Features/Feature.cpp
+++ b/rts/Sim/Features/Feature.cpp
@@ -399,9 +399,6 @@ void CFeature::DoDamage(
 	int weaponDefID,
 	int projectileID
 ) {
-	// paralyzers do not damage features
-	if (damages.paralyzeDamageTime)
-		return;
 	if (IsInVoid())
 		return;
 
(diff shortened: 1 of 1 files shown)


rtri:
?

Don't tell me ZK has frankenstein'ed combi-weapons that both damage and paralyze.

ashdnazg:
They did for the past few years.
Makes good gameplay too

sprunk:
ZK's issue can be solved by swapping the types (make the weapon deal normal damage and add extra paralyzer with Lua instead of the other way around).

However I think that it would still be elegant for the engine not to have such hardcoded limitations. If someone wants paralyzer not to damage wrecks they can replicate it with armorDefs or Lua.

After some thought this PR would need a bit more work first to have quality though: paralyzer needs to be passed to Feature[Pre]Damaged, this is a problem because either it gets appended last and becomes inconsistent with Unit[Pre]Damaged, or breaks backward compatibility.

GoogleFrog:
Is anyone using Feature[Pre]Damaged? The only issue I see here is backwards compatibility, the engine should be more general when the gameside change for compatibility is a little extra in armordefs.

rtri:
The proper generalization here will also affect Unit[Pre]Damaged and `CUnit` code that assumes damage is either/or.

sprunk:
How about now?

Features now call Feature[Pre]Damaged when getting damaged by paralyzer weapons but otherwise there is no change (damage type is still Xor, features still do not lose health when getting paralyzer damage).

carlosalfonso2014:
Thanks a lot!

PR #308, opened 2017-05-03
Fix streamCopy

rlcevg:
Only 1 AI gets AIData. Others don't because loadStream is invalid.


ashdnazg:
Something here doesn't make sense. Shouldn't Each AI get a separate block in the stream?

rtri:
They should, if AI's loaded their own saved state rather than the global game state.

rlcevg:
I also think that AI should load only own saved data: each AI needs own section.
Current fix is for streamCopy only as it looks like reuasble function.

PR #310, opened 2017-05-23
SetFeatureResurrect:

sprunk:
Previously: 2nd arg means unit def name if string, rez progress if number
Now: 2nd arg means unit def name if string, unit def id if number. Rez progress is 4th arg


rtri:
I have no objection to this, but just realized it should also be possible to un-set the resurrect target so `if (ud != nullptr)' can go.

PR #313, opened 2017-06-10
[AI] Only istream gets reset, not ostream

rlcevg:
ostream is an output, it doesn't seem logical to reset output in streamCopy (SkirmishAIWrapper.cpp)

P.S. Actually resetting istream after each AI started to use own load-stream is actually pointless with current usage. But still streamCopy, if it's really "copy", shouldn't change input in theory.


rlcevg:
I just realized that << and >> are for text only operations with stream, so CregLoadSaveHandler's AI part must be broken. Either each _size_ should end with std::endl, or read/write stream functions should be used.

PR #314, opened 2017-06-10
[AI] Only istream gets reset, not ostream. Safe aiSize for CregLoadSave

rlcevg:
ostream is an output, it doesn't seem logical to reset output in streamCopy (SkirmishAIWrapper.cpp)

Actually resetting istream after each AI started to use own load-stream is actually pointless with current usage. But still streamCopy, if it's really "copy", shouldn't change input in theory.

Edit: I should have probably mention earlier that this is not an enhancement, but a bug fix.
Currently each subsequent AI saves all previous data plus its own. Save/Load is broken.

### DIFF
--- a/rts/System/LoadSave/CregLoadSaveHandler.cpp
+++ b/rts/System/LoadSave/CregLoadSaveHandler.cpp
@@ -143,8 +143,7 @@ void CCregLoadSaveHandler::SaveGame(const std::string& path)
 			creg::COutputStreamSerializer os;
 			os.SavePackage(&oss, &gsc, gsc.GetClass());
 			PrintSize("Game", oss.tellp());
-		}
-		{
+
 			// save AI state
 			const int aiStart = oss.tellp();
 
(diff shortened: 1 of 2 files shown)


ashdnazg:
'\n' in a binary file seems a bit out of place
How about calling SerializeInt?

rlcevg:
I agree.
To my surprize ```oss << aiSize;``` writes aiSize in text mode (maybe it can be changed through stream configuration, don't know).
SerializeInt is the better way. How to do it? Or if you know, simply override/disregard my '\n' changes.

ashdnazg:
SerializeInt won't help as it's in the OutputStreamSerializer :S

rlcevg:
Why not? Simply combine 2 blocks ```// save creg state``` and ```// save AI state```, that way os (creg::COutputStreamSerializer) will be still reachable.

ashdnazg:
I haven't slept much tonight, seems to affect my creativity :)

PR #316, opened 2017-06-15
fix for d5f6e13

abma:
(CLIENT_NETLOG should be renamed maybe but i tried to keep the diff small so there will be fewer merge conflicts)


abma:
https://github.com/spring/spring/commit/28d81a9f9ff2f62d07c3cc92eb86a0fe8ed96fd5 is the better solution IMO

PR #317, opened 2017-06-23
make sure there is no nanoframe left when a factory is destructed

CommanderSpice:
This fixes units being built without a factory.


rtri:
If `StartBuild` can be called after `KillUnit`, checking for `isDead` at the top of it should achieve the same without spawning the frame.

CommanderSpice:
That is much cleaner. Updated.

PR #318, opened 2017-06-30
nullptr safety

sprunk:
`GetMouseStartPosition` has a null check but the other mouse funcs didn't.

### DIFF
--- a/rts/Lua/LuaUnsyncedRead.cpp
+++ b/rts/Lua/LuaUnsyncedRead.cpp
@@ -1890,6 +1890,9 @@ int LuaUnsyncedRead::GetActivePage(lua_State* L)
 
 int LuaUnsyncedRead::GetMouseState(lua_State* L)
 {
+	if (mouse == nullptr)
+		return 0;
+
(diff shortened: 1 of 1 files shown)


ashdnazg:
Actually, I think there's a mouse handler even in headless.
Maybe replace the ifs with an assert?

sprunk:
`GetMouseState` also dereferences `globalRendering` (and some other functions do, without any checks). How about like this?

```
assert (mouse != nullptr);
if (globalRendering != nullptr)
    return 0;
````

ashdnazg:
can globalrendering be nullptr while lua is running?

sprunk:
A null check does exist in one place: https://github.com/sprunk/spring/blob/a32563f8491a0fc88623f03ed511f220c72bd826/rts/Lua/LuaUnsyncedRead.cpp#L1725

However there are 10 other ones that do not have it so possibly it can't? I don't know.

rtri:
`globalRendering` is always non-null as far as all Lua states are concerned, any explicit checks either predate this or were erroneously added.

I second using asserts.

rtri:
closed by d7d5e660

PR #319, opened 2017-07-15
Fixing compilation errors when DEBUG_GLSTATE=ON

sanguinariojoe:
Actually I don't know if DEBUG_GLSTATE=ON is deprecated, or even if you are planning to deprecate it. Anyway, I guess this tiny fix will not hurt


ashdnazg:
thanks!

PR #320, opened 2017-07-17
PackCommandQueue: dont overallocate array

sprunk:
Prevents allocating a bigass table when getting just one command. Getting just the front of the command queue is a fairly common action.


sprunk:
Possibly there should just be a `GetQueueFront(unitID)` which returns just `cmd` (instead of having to  `GetCommandQueue(unitID, 1)` and getting `{ [1] = cmd }`)

rtri:
nice, thx

PR #324, opened 2017-07-19
Change GetUnitWeaponHaveFreeLineOfFire API

sprunk:
Possible calls currently are:
 * `(unitID, weaponNum, targetID)`
 * `(unitID, weaponNum, sourceX, sourceY, sourceZ, targetX, targetY, targetZ)`
 * `(unitID, weaponNum,     nil,     nil,     nil, targetX, targetY, targetZ)` required to specify target by XYZ but default source to aimpos

This has problems:
 * `(unitID, weaponNum, targetID, sourceX, sourceY, sourceZ)` is missing: using source by XYZ disallows target by ID, has to be XYZ as well
 * `(unitID, weaponNum, x, y, z)` specifies source and defaults target to (0,0,0) - hardly intuitive and breaks backwards compatibility
 * partial triplets such as `(unitID, weaponNum, x, y)` are accepted - prone to accidents

The PR puts target before source in all cases, allows to append source regardless of whether target is by ID or XYZ, and requires full triplets. Valid calls are now:

 * `(unitID, weaponNum, targetID)`
 * `(unitID, weaponNum, targetID, sourceX, sourceY, sourceZ)`
 * `(unitID, weaponNum, targetX, targetY, targetZ)`
 * `(unitID, weaponNum, targetX, targetY, targetZ, sourceX, sourceY, sourceZ)`

### DIFF
--- a/rts/Lua/LuaSyncedRead.cpp
+++ b/rts/Lua/LuaSyncedRead.cpp
@@ -3536,9 +3536,9 @@ int LuaSyncedRead::GetUnitWeaponHaveFreeLineOfFire(lua_State* L)
 
 		break;
 	case 2+3+0: // target XYZ, source default
-		tgtPos.x = luaL_optnumber(L, 3);
-		tgtPos.y = luaL_optnumber(L, 4);
-		tgtPos.z = luaL_optnumber(L, 5);
+		tgtPos.x = lua_tonumber(L, 3);
+		tgtPos.y = lua_tonumber(L, 4);
+		tgtPos.z = lua_tonumber(L, 5);
 		srcPos = weapon->GetAimFromPos();
 
 		break;
--- a/rts/Lua/LuaSyncedRead.cpp
+++ b/rts/Lua/LuaSyncedRead.cpp
@@ -3512,22 +3512,47 @@ int LuaSyncedRead::GetUnitWeaponHaveFreeLineOfFire(lua_State* L)
 	const CWeapon* weapon = unit->weapons[weaponNum];
 	const CUnit* enemy = nullptr;
 
-	float3 srcPos = weapon->GetAimFromPos();
+	float3 srcPos;
 	float3 tgtPos;
 
-	if (lua_gettop(L) <= 3) {
+	switch (lua_gettop(L)) {
+
+	case 2+1+0: // target ID; source default
 		if ((enemy = ParseUnit(L, __func__, 3)) == nullptr)
 			return 0;
 
 		tgtPos = weapon->GetUnitLeadTargetPos(enemy);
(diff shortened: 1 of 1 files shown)


ashdnazg:
Feature freeze is in effect, so unless it's crucial for something, I'd prefer not to merge this before 104.0, .

sprunk:
Is such a shuffle a "feature"? I worry that if the current API gets released then people would start using it, making future changes harder.

ashdnazg:
Devs are not expected to fix their code every RC.
Breaking backward compatibility without a very good reason before release is not a way to behave

sprunk:
The current API already breaks back comp from 103.0 (old API had args 3 4 5 as target XYZ, now they're source XYZ) and my proposed API is compatible with the old one (basically just adds source at the back of it instead of putting it inbetween).

ashdnazg:
You are correct, but what's the point of a feature freeze if no one can rely on it?
@rtri thoughts?

sprunk:
Is such a shuffle a "feature"? I consider it a bugfix (the main bug being "current api breaks back compat" and to some extent "current api has suboptimal design").

ashdnazg:
feature freeze isn't about "features" it's about making testing easier before the release. Obviously this includes not adding features as they may cause bugs, but that's not all. It should provide a stable interface to users so once they upgrade, they can test with us as bugfixes are being committed.

sprunk:
>making testing easier before the release (...) provide a stable interface to users

Surely restoring backwards compatibility does that, especially before the break is actually released?

rtri:
IMO some common sense has to be expected of game devs, i.e. that they don't pass partial triplets unless explicitly intended. Many of the API functions take this stance so as to not get bloated with error-handling logic for every possible permutation of arguments.

As it stands calling `GetUnitWeaponHaveFreeLineOfFire` with three args cause source and target to be taken from the weapon (as before) while four or more means *both* should be set manually. I don't see the problem there since 1) this callout is not widely used and 2) getting the old behavior back is as simple as inserting nils for the source parameters, i.e. `GetUnitWeaponHaveFreeLineOfFire(unitID, weaponNum, nil, nil, nil, tgtX, tgtY, tgtZ)`. Also, target before source is just fugly.

ashdnazg:
Since it's a thing that has an immediate use and that the current screen is hideous, it's acceptable imo, as long as it's not complex code.
Something like checking if a file of specific name from pwd exists and drawing it if it does - good.
Caching the splash screen from the last game ran by spring - bad.

sprunk:
Alright.

sprunk:
I've made a lite version of the proposal: #326

PR #325, opened 2017-07-21
Allow setting random seed in LuaParser

sprunk:
Using a constant seed can be useful for debugging.


rtri:
Not necessary, I intend to give `LuaParser` its own seedable generator.

PR #326, opened 2017-07-21
GetUnitWeaponHaveFreeLineOfFire allows target by ID

sprunk:
Lite version of previous. Allows to specify target by ID instead of XYZ but the rest stays the same.


sprunk:
f03115b594b04995aca4e13b8dccc456a2eb607c

PR #327, opened 2017-07-22
Fix Spring.GetSelectedUnitsSorted().n and Spring.GetSelectedUnitsCounts().n

CommanderSpice:
This fixes a bug introduced in 2dd3748ad4.
The "n" value in the tables returned to Lua always has the maximum value and is most likely wrong:
std::vector::resize() resizes the vector and std::vector::size() returns that size.

This patch counts the entries inserted into the table and stores the correct count in the table.


ashdnazg:
This PR is quite git mess, you may want to remake it over the actual develop.
do a force pull from origin, apply your changes and then do a force push to your repo
This will update the PR automatically

rtri:
cheers, I'll fix this manually.

PR #328, opened 2017-07-30
Air transport fixes

CommanderSpice:
I am not sure about the third one, Phoenix Annihilation works fine without the wait frames.

### DIFF
--- a/rts/Sim/MoveTypes/HoverAirMoveType.cpp
+++ b/rts/Sim/MoveTypes/HoverAirMoveType.cpp
@@ -404,11 +404,14 @@ void CHoverAirMoveType::UpdateHovering()
 	UpdateAirPhysics();
 	#endif
 
+	#if 1
 	// Random hovering is disabled because it breaks transport loading.
 	// TODO: Re-enable random wind.
-	#if 0
-	randomWind.x = randomWind.x * 0.9f + (gsRNG.NextFloat() - 0.5f) * 0.5f;
-	randomWind.z = randomWind.z * 0.9f + (gsRNG.NextFloat() - 0.5f) * 0.5f;
+	randomWind.x = 0.0f;
+	randomWind.z = 0.0f;
+
+//	randomWind.x = randomWind.x * 0.9f + (gsRNG.NextFloat() - 0.5f) * 0.5f;
+//	randomWind.z = randomWind.z * 0.9f + (gsRNG.NextFloat() - 0.5f) * 0.5f;
 
 	// randomly drift (but not too far from goal-position; a larger
 	// deviation causes a larger wantedSpeed back in its direction)
--- a/rts/Sim/MoveTypes/HoverAirMoveType.cpp
+++ b/rts/Sim/MoveTypes/HoverAirMoveType.cpp
@@ -404,7 +404,9 @@ void CHoverAirMoveType::UpdateHovering()
 	UpdateAirPhysics();
 	#endif
 
-	#if 1
+	// Random hovering is disabled because it breaks transport loading.
+	// TODO: Re-enable random wind.
+	#if 0
 	randomWind.x = randomWind.x * 0.9f + (gsRNG.NextFloat() - 0.5f) * 0.5f;
 	randomWind.z = randomWind.z * 0.9f + (gsRNG.NextFloat() - 0.5f) * 0.5f;
 
@@ -454,6 +456,8 @@ void CHoverAirMoveType::UpdateFlying()
 				const bool isTransporter = owner->unitDef->IsTransportUnit();
 				const bool hasLoadCmds = isTransporter && (topCmdID == CMD_LOAD_ONTO || topCmdID == CMD_LOAD_UNITS);
 				// [?] transport aircraft need some time to detect that they can pickup
+				// This looks like a dirty hack to give the transport a chance to hover near the unit to be loaded.
+				// TODO: Remove wait frames
 				const bool canLoad = isTransporter && (++waitCounter < ((GAME_SPEED << 1) - 5));
 				const bool isBusy = IsUnitBusy(owner);
 
@@ -465,6 +469,8 @@ void CHoverAirMoveType::UpdateFlying()
 							wantedHeight = orgWantedHeight;
 						}
 
+						// Hovering here can lead to goalPos never being reached
+						// TODO: Remove hovering state before transport loading
 						SetState(AIRCRAFT_HOVERING);
 					} else {
 						if (!isBusy) {


rts/Sim/MoveTypes/HoverAirMoveType.cpp
```diff
@@ -410,7 +410,7 @@ void CHoverAirMoveType::UpdateHovering()
 
 	// randomly drift (but not too far from goal-position)
 	wantedSpeed = (randomWind * math::fabs(owner->unitDef->dlHoverFactor) * 0.5f);
-	wantedSpeed += (smoothstep(0.0f, 20.0f * 20.0f, (goalPos - owner->pos)) * (goalPos - owner->pos));
+	wantedSpeed += (goalPos - owner->pos) * 0.1f;
```
ashdnazg: While smoothstep was indeed wrong here because it doesn't take into account where the movement started, this solution is bad as well and can easily cause overshoot.

rts/Sim/MoveTypes/HoverAirMoveType.cpp
```diff
@@ -463,8 +463,6 @@ void CHoverAirMoveType::UpdateFlying()
 						if (waitCounter > (GAME_SPEED << 1)) {
 							wantedHeight = orgWantedHeight;
 						}
-
-						SetState(AIRCRAFT_HOVERING);
```
ashdnazg: I am very very suspicious about this.

rts/Sim/MoveTypes/HoverAirMoveType.cpp
```diff
@@ -452,17 +452,13 @@ void CHoverAirMoveType::UpdateFlying()
...
-				if (!CanLand(isBusy) || (canLoad && hasLoadCmds)) {
+				if (!CanLand(isBusy) || hasLoadCmds) {
 					wantedSpeed = ZeroVector;
 
 					if (isTransporter) {
-						if (waitCounter > (GAME_SPEED << 1)) {
-							wantedHeight = orgWantedHeight;
-						}
+						wantedHeight = orgWantedHeight;
 					} else {
```
ashdnazg: this one is probably ok I guess?

rts/Sim/MoveTypes/HoverAirMoveType.cpp
ashdnazg: While smoothstep was indeed wrong here because it doesn't take into account where the movement started, this solution is bad as well and can easily cause overshoot.

rts/Sim/MoveTypes/HoverAirMoveType.cpp
```diff
@@ -452,19 +452,13 @@ void CHoverAirMoveType::UpdateFlying()
...
-				if (!CanLand(isBusy) || (canLoad && hasLoadCmds)) {
+				if (!CanLand(isBusy) || hasLoadCmds) {
 					wantedSpeed = ZeroVector;
 
 					if (isTransporter) {
-						if (waitCounter > (GAME_SPEED << 1)) {
-							wantedHeight = orgWantedHeight;
-						}
-
-						SetState(AIRCRAFT_HOVERING);
```
ashdnazg: I am very suspicious about removing ```SetState(AIRCRAFT_HOVERING)```

rts/Sim/MoveTypes/HoverAirMoveType.cpp
CommanderSpice: It just causes the random wind to blow it a bit more towards goalPos than away from it.
How do you think this can overshoot?

CommanderSpice: Ok, I see. If the randomWind would blow it past goalPos, this nudges it even further past it.

The problem with the smoothstep is that it applies a drift into quadrant I. It never pushes the unit back into -X-Y.

rts/Sim/MoveTypes/HoverAirMoveType.cpp
CommanderSpice: You are right to be suspicious. The overshooting it fixes comes from my changes to UpdateHovering. :/

rts/Sim/MoveTypes/HoverAirMoveType.cpp
ashdnazg: it actually should, and the problem is different I think.

rts/Sim/MoveTypes/HoverAirMoveType.cpp
ashdnazg: I'll try fixing the hovering tomorrow

rts/Sim/MoveTypes/HoverAirMoveType.cpp
CommanderSpice: But it doesn't. `smoothstep` interpolates between 0,0,0 and the magic 200,200,200.
That is a drift into +X+Y.

rts/Sim/MoveTypes/HoverAirMoveType.cpp
CommanderSpice: It occured to me that for hovering to make sense here, the hovering has to guarantee that it will cross goalPos soon. And as the hovering is random, it can't.

CommanderSpice:
I reverted the last commit. It does not fix anything and just distracts.

ashdnazg:
I think @rtri did https://github.com/spring/spring/commit/70d14a0623815a9b9b8797bc68b4b8b67ec9d0ac as an alternative to this PR

CommanderSpice:
That fix does not fix it.

CommanderSpice:
Also, the broken hovering is just one half of this PR. The other half is using hovering when there is a pickup order.

rtri:
I was looking into 4827 again when this PR arrived (appearing to describe the same issue) and wanted to apply the minimally-invasive fix with 104 in mind.

If the hovering state is also active during pickup that explains the `waitCounter` and randomizing `wantedSpeed` should simply be disabled in that case.

CommanderSpice:
So you want to just disable hovering to fix loading until after the release?

CommanderSpice:
I think hovering is used from closeToGoal until it reaches goalVec.
Then (if that ever happens) it switches to landing.

CommanderSpice:
Actually, this is worse. They take a very long time to reach goalPos.

rtri:
It's disabled now, but I could not reproduce the problems you are seeing ("they take a very long time to reach goalPos") in PA. The scenarios that did fail in my tests (e.g. armatlas not picking up armflash) are instances of https://springrts.com/phpbb/viewtopic.php?f=14&t=35808 afaict.

CommanderSpice:
It's fixed. Thanks!

PR #329, opened 2017-07-31
Disallow building dead units

sprunk:
To prevent the case where UnitFinished happens after UnitDestroyed.


ashdnazg:
If this can happen we may want to prevent the builders from trying to build/repair it as well (so they can do other things for instance)

sprunk:
Happens just for 1 frame as far as I can tell

ashdnazg:
crashing units though

sprunk:
Right. It would be a good thing to add but I don't think I'd be able to do it soon.

ashdnazg:
thanks

PR #330, opened 2017-08-01
CallAsTeam: full access only if AllAccessTeam

sprunk:
Currently CallAsTeam does not temporarily cancel full read access. This means that, for example, GetUnitPosition returns the real position even if there's radar wobble (which means it's impossible to get the actual wobbled position because GetUnitPosErrorParams correctly returns nil, since the unit is enemy!)


ashdnazg:
nice catch, thanks!

PR #331, opened 2017-08-08
GetGrass: false/true bool instead of 0/1 int

sprunk:
0 and 1 are both true in Lua which makes 0/1 annoying.


gajop:
This shouldn't be merged.
Grass should ideally be represented in a [0, 1] or [0, 255] range, representing thickness.

It's the `Spring.{Add,Remove}Grass` interface that needs changing, but I didn't want to do it now.

sprunk:
Makes sense.

gajop:
Actually, I was also thinking grass shouldn't really be part of synced anyway.
Since there was no way to read it, it never really influenced simulation. 

It might be best to make it completely unsynced, in a texture, but that might be difficult to implement (not sure).

PR #332, opened 2017-08-14
[AI] Make IsNeutral available in los or radar

rlcevg:
Currently IsNeautral can be read only if unit is in los. I'm unsure about the original purpose of neutral flag;  this change allows to read its value in IN_LOS or IN_RADAR (sonar is IN_RADAR, right?)

related: https://github.com/rlcevg/CircuitAI/issues/59


rtri:
lgtm, ty

PR #333, opened 2017-08-22
fix SetDecalAlpha

sprunk:


ashdnazg:
ty

gajop:
Wait a minute, does decalsGl4 actually work? I thought it was unused

ashdnazg:
I'm not sure, but the error was an error nevertheless

gajop:
I agree of course, but due to https://github.com/sprunk/spring/blob/d17d18bf616aec93d847a9cecf50df3f97fb20e5/rts/Rendering/Env/IGroundDecalDrawer.cpp#L27-L36 I think it is disabled. Was wondering if someone was using it and ran into this problem or not, @sprunk ?

sprunk:
No, I just saw it in the code.

gajop:
Okey, thanks

PR #336, opened 2017-08-24
Remove legacy hack supposed to die in 90.0

sprunk:


rtri:
The majority of games still rely on COB, but that does not legitimize this (forgotten) hack. Feature-freeze is however not the time to be removing any.

PR #337, opened 2017-09-03
Add macos buildslave

TurBoss:
this modifies the master.cfg file of the buildbot and adds a cmake file with the cmake options


Anarchid:
Glorious victory awaits.

PR #339, opened 2017-09-20
Spring.GetPlayerList:

sprunk:
* accepts AllAccessTeam to list everybody (default when no arg given)
 * accepts NoAccessTeam to only list specs
 * teamID == 0 no longer lists specs (in addition to players actually on team 0)

### DIFF
--- a/rts/Lua/LuaSyncedRead.cpp
+++ b/rts/Lua/LuaSyncedRead.cpp
@@ -1084,7 +1084,7 @@ int LuaSyncedRead::GetTeamList(lua_State* L)
 
 int LuaSyncedRead::GetPlayerList(lua_State* L)
 {
-	int teamID = -1;
+	int teamID = CEventClient::AllAccessTeam;
 	bool active = false;
 
 	if (lua_isnumber(L, 1)) {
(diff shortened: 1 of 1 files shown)


ashdnazg:
I'm not sure if I like the use of {No,All}AccessTeam as these are used for visibility.
Why should specs be NoAccessTeam when they usually have AllAccessTeam visibility?

sprunk:
They are also used for "control", just there are no Lua callouts (yet) that deal with control. (random example: https://github.com/spring/spring/blob/bfbe8c2f3eb21ec385d8f3998c94659badb43500/rts/Lua/LuaUtils.h#L286 )

ashdnazg:
I see, on that case, using NoAccessTeam makes sense, but AllAccessTeam wouldn't make sense for getting all players

sprunk:
Fair enough. How about a 3rd boolean parameter for `GetPlayerList` that filters specs out (defaulting false to keep current behaviour) and a new `GetSpectatorList` callout (to do what NoAccessTeam did)?

ashdnazg:
Nah, the teamID==0 behaviour should definitely be dropped.

sprunk:
Specs should probably be -1 but then you need to review the codebase to make sure teamID is never used as an array index.

Gaia makes a lot of sense for spectators too, especially since it cannot ever have players on it.

ashdnazg:
Definitely not Gaia. specs don't control Gaia - period

PR #341, opened 2017-09-22
Allow resurrecting indestructible features

sprunk:


ashdnazg:
I wonder what was the reason for not allowing indestructible features to be rezed.

sprunk:
It makes some sense because rezzing a feature "destroys" it in a way. However you can just make a feature unrezzable if you want to.  I think the check comes from before the era of Lua gadgetry where you couldn't.

rtri:
That, and/or indestructible features were assumed to be part of the map scenery which rezzers should not waste time on.

PR #342, opened 2017-09-25
DefaultCommand call-in:

sprunk:
Previously:
 * widgets run the callin "blind" (they dont know what the engine command would be)
 * unsynced gadgets are run only if widgets dont return anything (they implicitly know widgets didnt override but dont know the engine command either)

Now:
 * unsynced gadgets are run first, they receive the engine default cmd as 3rd param
 * widgets are then run and receive current default (taking into account what gadgets said)


ashdnazg:
ty

sprunk:
t**y**

PR #344, opened 2017-09-30
cleanup of copy elision warnings

dzosz:
fix perf warnings reported by clang 4.0.1
`warning: moving a temporary object prevents copy elision [-Wpessimizing-move]`


rtri:
Thanks.

However, we are explicit about generating moves as a matter of style so we don't have to worry about compilers missing c.e. opportunities. Can you put the `override` diffs in a separate PR?

dzosz:
hey, my understanding was that explicit std::move in some cases will actually prevent possible rvo/copy elision (as suggested by compiler warning), not the other way around. In most cases we're asking for value returned by copy:
`const std::string soundExt = std::move(FileSystem::GetExtension(fileName));
`
using std::move seemed counter intuitive. It won't avoid the copy and potentially (?) adds overhead.

rtri:
> my understanding was that explicit std::move in some cases will actually prevent possible rvo/copy elision

It will, the issue that c.e / rvo is not guaranteed to take place. With explicit moves it does not matter what the compiler would choose to do.

ashdnazg:
it may be interesting to either profile this or check the generated assembly and see the effects of optimisation or lack thereof.

dzosz:
I benchmarked construction of four different structs with explicit moves vs no move marking.

Results based on 3 runs:
- Struct with strings 3% penalty
- PODs 40% penalty
- 3x char arrays[20] 53% penalty
- Mixed struct 43% penalty

raw data:
https://gist.github.com/dzosz/d8c7d5038927c9dfbb577eeadcb01ffa
source code for benchmark:
https://github.com/dzosz/cpp_move_penalty_benchmark

gajop:
I think we should deal with this PR, it's been a year.
Some cases of copy elision have been removed from the engine, but a few still remain (quick grep):

```
Game/ClientData.cpp:32: return std::move(zlib::deflate(reinterpret_cast<const std::uint8_t*>(clientData.data()), clientData.size()));
Game/ClientData.cpp:40: return (std::move(cdata));
Game/UI/MiniMap.cpp:808:                return (std::move(CTooltipConsole::MakeUnitString(unit)));
Game/UI/MiniMap.cpp:814:        return (std::move(CTooltipConsole::MakeGroundString({wpos.x, CGround::GetHeightReal(wpos.x, wpos.z, false), wpos.z})));
Rendering/Shaders/Shader.cpp:118:       return (std::move(version));
Rendering/Shaders/Shader.cpp:551:               return (std::move(cso));
System/EventHandler.cpp:677:            return (std::move(str));
System/StringUtil.cpp:262:      return (std::move(deflData));
System/StringUtil.cpp:289:      return (std::move(inflData));
```

I'd recommend getting rid of them all (maybe I should make a new PR?). `return std::move(...)` does nothing besides preventing copy elision (which is more efficient than move). A move will still happen if copy elision is not possible! `std::move` doesn't move, it just casts it into an xvalue and makes moves possible. Whether they happen or not depends on the existence of a move constructor.

There are plenty of references (https://en.cppreference.com/w/cpp/utility/move , https://en.cppreference.com/w/cpp/language/copy_elision), but I think this SO answer is easiest to understand: https://stackoverflow.com/questions/4986673/c11-rvalues-and-move-semantics-confusion-return-statement?answertab=votes#tab-top (not a reference, but comes from the guy who essentially designed move semantics so I'd say we can take it for granted).

PR #345, opened 2017-10-02
Fix rendering of minimap for DualScreenMode=1 and MiniMapRenderToTexture=1

jrb0001:


rtri:
thanks for the patch, some copypastaccident apparently happened here.

PR #347, opened 2017-10-05
remove reference to moved object

dzosz:
Object _model_ is first moved out, but after that we reference to its member _id_ which **should** be in unspecified state. Fortunately move assignment did not apply move to all of its members. I left _explicit moves_ as they were before, as this is deserves separate discussion.


rtri:
thanks but this is unnecessary: `id` is a simple integer, moving a model just copies it.

if you want, you could change `return &models[model.id];` to `return &models[models.size() - 1];` instead.

dzosz:
OK I see. PODs are faster to copy. Though, I start to wonder whether compiler can apply better optimisations when all of the members are moved .

PR #348, opened 2017-10-05
`/specfullview` calls unsynced `PlayerChanged`.

sprunk:


rtri:
Hmm, what's the added benefit over PlayerChanged + Spring.GetSpectatingState?

sprunk:
`PlayerChanged` not called when doing `/specfullview x`.

rts/Game/UnsyncedGameCommands.cpp
```diff
@@ -934,8 +934,8 @@ class SpecFullViewActionExecutor : public IUnsyncedActionExecutor {
 			gu->spectatingFullSelect = gu->spectatingFullView;
 		}
 		CLuaUI::UpdateTeams();
-		// NOTE: unsynced, so do not inform via eventHandler
-		unitDrawer->PlayerChanged(gu->myPlayerNum);
+		eventHandler.SpecFullViewChanged(gu->spectatingFullView, gu->spectatingFullSelect);
+		unitDrawer->PlayerChanged(gu->myPlayerNum); // NOTE: unsynced, so do not inform via eventHandler
```
sprunk: Note that this does not go through `eventHandler` so Lua won't know.

sprunk: After some thought it might be better to just do

`unsyncedLuaHandle->PlayerChanged(gu->myPlayerNum)`

here. A bit inelegant skipping the handler but I guess more elegant than an extra call-in?

rtri: You can also make /specfullview send something over the network, shouldn't be many more lines.

ashdnazg: far less clean IMO

sprunk: >far less clean IMO
The "skip EH and notify unsynced directly" or "send over network"?

As far as I see, sending it over the network has issues:

 * if I do the `gu->spectatingFullFoo = x` in unsynced and just use the synced part for notification, then there's the network-lag period of time where widgets don't yet know that the state changed and might be showing more/less than the player wants.
 * if I do that part in synced there's unnecessary extra lag before the change actually takes place.
 * synced `PlayerChanged` will probably desync replays if a game happens to have logic there?

I think I prefer notifying unsynced lua handles directly (skipping EH).

ashdnazg: The "send over network" is very unclean.

Maybe make the engine SpecFullView event call the PlayerChanged callin so there's no need for 20 callins?

sprunk:
As far as I see, sending it over the network has issues:

 * if I do the `gu->spectatingFullFoo = x` in unsynced and just use the synced part for notification, then there's the network-lag period of time where widgets don't yet know that the state changed and might be showing more/less than the player wants.
 * if I do that part in synced there's unnecessary extra lag before the change actually takes place.
 * synced `PlayerChanged` will probably desync replays if a game happens to have sync-affecting logic there?

I think I prefer notifying unsynced lua handles directly (skipping EH) as the least inelegant.

ashdnazg:
good enough for me I guess...

thanks!

PR #349, opened 2017-10-08
Allow offsetting tracking lights

jrb0001:


rts/Sim/Projectiles/Projectile.cpp
```diff
@@ -212,7 +212,7 @@ bool CProjectile::GetMemberInfo(SExpGenSpawnableMemberInfo& memberInfo)
 	if (CExpGenSpawnable::GetMemberInfo(memberInfo))
 		return true;
 
-	CHECK_MEMBER_INFO_FLOAT3(CProjectile, dir)
+	//CHECK_MEMBER_INFO_FLOAT3(CProjectile, dir)
```
jrb0001: I don't know spring good enough to find the best way to solve this. Please let me know how you would do it.

ashdnazg:
Can you please give a concise explanation of what you are trying to do?

ashdnazg:
In general I recommend discussing changes in the forums/#sy before pull requests, especially when you "don't know spring good enough".

rtri:
From the code, it looks like the intention is to offset (=translate) tracking lights within a tracked object's frame but this might not be the best way to get there.

ashdnazg:
I'm still not sure where the translation is supposed to come from

rtri:
`light.GetPosition()` probably, interpreted as a vector relative to `light.GetTrackPosition()`

jrb0001:
Yes, the idea is to translate and rotate the lights by their own position and rotation relative to the position and rotation of the unit or projectile which they are tracking.

PR #350, opened 2017-10-13
Cheat commands interface:

sprunk:
* `/nocost` accepts optional bool parameter instead of always toggling
 * add `Spring.GetGlobalLos(allyTeamID)`
 * add `Spring.IsNoCostEnabled`


ashdnazg:
ty

PR #351, opened 2017-10-30
Remove movedef refcounts

sprunk:
Currently movedefs keep refcounts based on unitdefs that use them. Any movedef that has 0 refs is then skipped when processing. This is bad:
 * you can't use `Spring.MoveCtrl.SetMoveDef` for those movedefs.
 * there's no point keeping such inaccessible movedefs in memory, wasteful.

Ergo:
 * games should be able to explicitly specify which movedefs they are going to use.
 * useless movedefs shouldn't even reach the engine.

This patch removes the refcounts; all movedefs provided by the game will be processed. It is now game's responsibility to say which movedefs it wants. For backward compatibility, the builtin defs parser now filters out unused movedefs.

### DIFF
--- a/cont/base/springcontent/gamedata/defs.lua
+++ b/cont/base/springcontent/gamedata/defs.lua
@@ -41,6 +41,8 @@ Spring.TimeCheck('[defs.lua] loading all *Defs tables:', function()
   DEFS.weaponDefs  = LoadDefs('weaponDefs')
   DEFS.armorDefs   = LoadDefs('armorDefs')
   DEFS.moveDefs    = LoadDefs('moveDefs')
+
+  LoadDefs('defs_post') -- misc cross-def postprocessing
 end)
 
 
--- a/cont/base/springcontent/gamedata/defs_post.lua
+++ b/cont/base/springcontent/gamedata/defs_post.lua
@@ -1,0 +1,22 @@
+-- filter out unused movedefs
+
+local usedMovedefs = {}
+for name, def in pairs (DEFS.unitDefs) do
+	local movedef = def.movementclass
+	if movedef then
+		usedMovedefs[movedef] = true
+	end
+end
+
+local i = 1
+while i <= #DEFS.moveDefs do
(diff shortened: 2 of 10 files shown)


ashdnazg:
Looks mostly good, please add the existence of defs_post.lua to the changelog

rtri:
> games should be able to explicitly specify which movedefs they are going to use.
 > useless movedefs shouldn't even reach the engine.

if you want to meet the second criterion (I do), then filtering still has to be done in-engine since a game can override `defs.lua`.

sprunk:
>>useless movedefs shouldn't even reach the engine.
>
>filtering still has to be handled in-engine since a game can override `defs.lua`.

That's good though, overriding `defs.lua` and letting in "unused" movedefs is desirable because that's how you tell the engine you're actually going to use them in `Spring.MoveCtrl.SetMoveDef`.

Personally I would've went a step further and not have created the `defs.lua` default at all because as-is the patch doesn't solve the `SetMoveDef` inelegance where it doesn't work out of the box with some movedefs and requires overriding some magic files.

rtri:
It ain't so desirable because I see a reality populated by "unused except by `SetMoveDef`" movedefs emerging; performance is a concern here.

ashdnazg:
I think game should be able to control which movedefs exist, since for a game, the alternative is making a fake unitdef just to force the engine to use a movedef.

In this PR (Disclosure: I kinda told Sprung what to do here) we still create a smart default that keeps the current behaviour, and which most devs will use as is.
If devs goes out of their way to create performance issues for themselves, this is their issue, not ours.

btw @sprunk 
1) you can add a fake tag to prevent your _post from removing said movedef.
2) By definition, if it's a lua code file, it's the exact opposite of a "magic file" - it's fully customisable.

abma:
@sprunk 

FYI: https://github.com/spring/spring/commit/b930b9d2499b96e55955dfe0f93ab979932190a2

PR #352, opened 2017-11-10
Kill LuaUI before clearing game rules params

sprunk:
LuaUI might want to access GRP on shutdown.


rts/Game/Game.cpp
```diff
@@ -814,9 +818,6 @@ void CGame::KillLua(bool dtor)
 	CLuaHandleSynced::ClearGameParams();
```
sprunk: GRP clear is here.

ashdnazg:
ty

PR #357, opened 2017-12-08
Allow setting `noAutoTarget` weapon def tag

sprunk:
Similar but different to manualfire (uses regular CMD.ATTACK, will fire multiple times instead of just once, etc)


sprunk:
eh probably not a good idea

PR #360, opened 2017-12-23
Sound queue now actually a queue (was stack)

sprunk:


ashdnazg:
maybe change streamQueue into deque and use pop_front?

ashdnazg:
thanks!

PR #362, opened 2018-01-31
Avoid deprecated socket_base::non_blocking_io

Mailaender:
http://www.boost.org/doc/libs/1_47_0/doc/html/boost_asio/reference/socket_base/non_blocking_io.html says it was deprecated since boost 1.47 (released 11.7.2011) and we are having trouble building this for @openSUSE Tumbleweed.

### DIFF
--- a/rts/System/Net/UDPListener.cpp
+++ b/rts/System/Net/UDPListener.cpp
@@ -32,8 +32,7 @@ UDPListener::UDPListener(int port, const std::string& ip): acceptNewConnections(
 	const std::string err = TryBindSocket(port, &socket, ip);
 
 	if (err.empty()) {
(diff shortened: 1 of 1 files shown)


rtri:
PR itself is fine, but I'm a little worried about the implicit suggestion that SUSE packages are built from Spring's development branch. Stable releases (currently 104.0) shoule be based on master, merging this will not immediately benefit you.

Mailaender:
We are not building your development branch. It is just good practice to upstream patches and I did it for https://build.opensuse.org/package/view_file/home:simonpuchert:branches:games/springrts/fix_deprecated.patch See https://build.opensuse.org/package/show/games/springrts for the discussion on our springrts package.

rtri:
alright, good to know

PR #364, opened 2018-02-05
LuaHandler: `pcall` when loading config

sprunk:


gajop:
Can it actually return a non-table if it doesn't error?

gajop:
Seems to me this should be put in a function then. Especially since the second example you've given seems to not print any error if you don't return a table (just silently fails).

PR #365, opened 2018-02-06
LaserCannon: use dynamic projectile speed

sprunk:


ashdnazg:
I have some vague memory of leaving it this way when I went over this code due to ttl stuff.

sprunk:
TTL should use the dynamic value because it affects range. If you halve projectile speed while keeping TTL you effectively halve range as well but the weapon doesn't know that and attempts to shoot anyway.

ashdnazg:
as I said, vague memory, will have to check that more thoroughly, that's all

PR #369, opened 2018-03-12
Add `SyncedPlayerChanged`

sprunk:
Except for demo-watching specs, when these do something it's still unsynced only


rtri:
This is moving in the wrong direction for my taste, I'd add a SyncedPlayerChanged event.

sprunk:
Alright.

abma:
@sprunk:

as i only partly understand this change: what problem does this solve?

sprunk:
Synced currently doesn't know about the PlayerChanged event, so you can't do things when somebody disconnects or reconnects (for example a gadget that gives the abandoned units so somebody can control them, and then returns them on reconnect). I believe it is unavailable because people watching a replay can also generate the event, which is why the patch makes a new callin (cutting replay-watching spectators from regular PlayerChanged would hurt unsynced gadgets and LuaUI, and exposing them to synced would open up the possibility of a desync).

abma:
ugh, the merge is wrong. i'll do in cli.

thanks for explaining!

sprunk:
I copied the explanation to the code as a comment: 747b49cd6bedc672283b3fd691b6e8143e8cd5bb

rts/Net/NetCommands.cpp
```diff
@@ -1291,7 +1291,20 @@ void CGame::ClientReadNet()
 						skirmishAIHandler.RemoveSkirmishAI(aiNum);
 
 						CPlayer::UpdateControlledTeams();
-						eventHandler.PlayerChanged(playerNum);
+						playerHandler.Player(playerNum)->NotifyPlayerChanged();
+
+						LOG("Skirmish AI \"%s\" (ID:%i), which controlled team %i is now dead", aiInstanceName.c_str(), aiNum, aiTeamId);
```
abma, line 1296: rts/Net/NetCommands.cpp:1296:79: error: use of undeclared identifier 'aiInstanceName'

  ...AI \"%s\" (ID:%i), which controlled team %i is now dead", aiInstanceName...

doc/changelog.txt
```diff
@@ -150,6 +151,8 @@ Lua:
  - MarkerAdd{Point,Line} now have an extra optional argument (after localOnly: 6th for point, 8th for line):
    the playerID of the player who should appear as the author of the drawing (if missing, defaults to the local player).
    Only has an effect when localOnly is true.
+
+=======
```
abma, line 155: my fault: should be removed

abma (changes requested):

abma:
https://github.com/spring/spring/commit/aa7a9972a26c15cb41d9d438fc457a86d58f7630

PR #376, opened 2018-05-08
make some opengl extensions optional:

abma:
allows to start spring with mesa drivers / opengl 4.5 (core profile)

without these changes spring can't be started...


rtri:
yes, all of these moved into core or were dropped.

PR #377, opened 2018-05-29
Adds lua AI version

tomjn:
Sometimes its useful to know this or have some idea of it, and who knows, a mod might want multiple versions and have a means of handling it


tomjn:
@ashdnazg fixed

PR #378, opened 2018-05-31
 Drawings callouts: allow to specify playerID

sprunk:
MarkerAdd{Point,Line} now have an extra optional argument (after localOnly: 6th for point, 8th for line) the playerID of the player who should appear as the author of the drawing (if missing, defaults to the local player). Only has an effect when localOnly is true.


rtri:
This is the last PR I'll handle on ZK's behalf for a while.

PR #381, opened 2018-06-06
GetPlayerInfo: restore backward compatibility

sprunk:
All code that uses player options was broken. This patch puts them back to 10th spot, putting the new qol thingie as the 11th.


rtri:
nah.

PR #383, opened 2018-06-12
Fix to 'invalid parameter array size' GLSL error

sanguinariojoe:
Apparently SMF render was crashing because the number of vec4 uniforms was not correctly computed


rtri:
ty

PR #387, opened 2018-07-11
A few goodies for OpenGL shaders:

lhog:
* Added support for Lua tesselation shaders:
	- New shaders TES & TCS
	- New function SetTesselationShaderParameter() to manipulate glPatchParameter*()
	- Renamed SetShaderParameter to SetGeometryShaderParameter to avoid confusion between Tesselation shaders and Geometry shader
	- New drawing primitive GL.PATCHES
	- TODO. LuaShaderContainer seems to duplicate LuaShaders code, so TES & TCS should go there too.

I've tested TES & TCS and they work with the following definitions in the tesselation shaders:
	#version 150 compatibility
	#extension GL_ARB_tessellation_shader : enable

* Also added support for subroutines. They are useful so different draw passes/techniques could be specified as subroutine uniforms.
  Currently one has to have several slightly different instances of the the shader to do so.
	- This code has not been tested yet.


gajop:
Changing the base from this GUI does pretty evil stuff ^_^
Best close this PR and open a new one when you're ready?

lhog:
@ashdnazg I'm not sure I have got the true meaning of your words:

Does this mean maintenance will never get what I've just authored or does it rather mean that it should go thru develop first?

If it's former I will be very disappointed, because no mod can switch develop in the reasonable amount of time nor develop is ready for production (AFAIK).
If - latter, I am a bit concerned too. The PR diff might not be portable between branches.

gajop:
All commits must go to develop, and only select ones should be cherry picked for maintenance.
Since we're moving to core profile, GFX stuff is unlikely to be cherry picked for maintenance.

I do get your concern about the switch, it's not going to be easy, but that's just the way things are.

lhog:
Based on that @gajop wrote, I'll make my question straight:

For the core engine team: @ashdnazg @rtri :
Provided I fix compilation and (possible) execution bugs, will this code eventually (through the develop branch) get to maintenance. Yes or no?

ashdnazg:
If this code goes through develop, it could be merged into maintenance.
I don't think it goes with the current direction of develop though, so that's up to @rtri .

lhog:
Out of curiosity.... I've taken a look at Lua's CreateShader() and it look exactly or almost exactly the same. Are we planning to get rid of/refactor CreateShader()? Otherwise I'm not sure what the following means:

> I don't think it goes with the current direction of develop though, so that's up to @rtri .

ashdnazg:
I won't be surprised if it will be changed in some way.
Again, no idea. Ask @rtri

lhog:
The headless compilation works for me locally, but fails here.
Any idea why the difference?

Anarchid:
> Any idea why the difference?

The log may help: https://travis-ci.org/spring/spring/jobs/404119235

lhog:
It can't help, because I cannot reproduce it locally having exact copy of the code. Maybe some 3rd party libs are different/outdated on the buildbot side?

abma:
the constans are not defined in https://github.com/spring/spring/tree/develop/include/GL *.h

i'm pretty sure you can reproduce it when you compile the target "spring-headless"

lhog:
Ok, now it makes some sense.
Locally on my end those constants come from https://github.com/spring/mingwlibs/tree/master/include/GL so I can build both engine and headless. Why does https://github.com/spring/spring/tree/develop/include/GL have so outdated GLEW?

lhog:
N.B. if this PR is to ever make into maintenance, it will need updated GL headers. MinGW ones seem to do the trick.

PR #388, opened 2018-07-19
A few goodies for OpenGL shaders (for develop)

lhog:
See the story here: https://github.com/spring/spring/pull/387

I recommend separate PRs are used for develop and maintenance (this one and https://github.com/spring/spring/pull/387 respectively), because they have slightly different changes.


lhog:
So any interest in this?

On the side note I'm actually surprised, current engine developers don't embrace the rare contributors who would like to help. You think one man can do everything himself? This is the path to neverland, not to mention the bus factor.
Hope I'm totally wrong and @rtri was just busy.

ashdnazg:
Yes, the bus factor sucks. Don't take it personally though. I'm pretty sure that if Kloot had the time to go through this, he would, and that he'd respond once he can.

lhog:
Thanks @gajop.
I'm in US this week, will get it sorted out this weekend or next week.

rtri:
1) There are millions of valid possible reasons beyond being busy why someone might drop work on a volunteer project without prior explanation or permission, one of them as simple as "it's summer".
2) If you feel offended by a PR not getting embraced immediately, maybe you can also understand why participating in an environment where the two major surviving game communities have chosen to ignore / reject new engine maintenance builds (thereby nullifying all time and effort sunk into them) isn't very interesting or rewarding anymore.

> You think one man can do everything himself?

One man might when all he has to do is drive a nearly empty bus around, but that is neither his idea of fun nor sustainable. It has also become obvious that development can only proceed in conjunction with regular testing and short feedback cycles, which are no longer happening with ZK frozen at 287-gf7b0fcc and BA back on 103.0, so said bus has been mostly parked until this situation changes.

PR itself looks good since LuaShaders probably won't see much renovation, thanks.

lhog:
Thank you @rtri and I'm sorry for what might have looked like an emotional outbreak.
It was never my intent to offend anyone. Guess I have chosen wrong words.

Can you also merge https://github.com/spring/spring/pull/387 ?

P.S. ZK is not only stuck at 287-gf7b0fcc, but look stagnant development wise for quite some time. Though it's definitely not a summer in Australia.

PR #389, opened 2018-07-20
Fix inconsistent nano buildrange

sprunk:
Nanos currently use a different buildrange equation than regular builders: they require to reach the center of the target, not the edge of its radius (except a very rare edge case where they are *inside* the radius in which case they are automatically in range)

### DIFF
--- a/rts/Sim/Units/CommandAI/BuilderCAI.cpp
+++ b/rts/Sim/Units/CommandAI/BuilderCAI.cpp
@@ -251,14 +251,6 @@ void CBuilderCAI::PostLoad()
 
 inline float CBuilderCAI::GetBuildRange(const float targetRadius) const
 {
-	// for immobile:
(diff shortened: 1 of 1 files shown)


rtri:
The difference apparently existed after attempted removal of a TA-ism in https://github.com/spring/spring/commit/b38c1546240dd13dc50f3f7d60c0afd1208139ab which wasn't entirely successful according to https://github.com/spring/spring/commit/538464d0c716a538f9722a73aa6696c5430d71b9.

If the "nanos assist stuff outside range" bug isn't reintroduced this can go in.

DeinFreund:
"nanos assist stuff outside range" is exactly what's happening right now. Nanos that should have 2D build range are trying to assist things within their 2D range, but fail with things that have a high vertical difference.

rtri:
Yes, I know. I don't see how that invalidates remarks about this (still untested) PR possibly causing a *separate* regression.

PR #391, opened 2018-07-31
add. be14d9c2ecbbd9fb9cacc56c02442164824e9071

sprunk:
Replace a magic number (previously it made some sense because https://github.com/spring/spring/commit/be14d9c2ecbbd9fb9cacc56c02442164824e9071#diff-ed40a10d6489adbd290ee8870e244f93L272 was defined in terms of slowupdate but that is no longer the case)


ashdnazg:
ty

PR #392, opened 2018-08-11
Unit[Pre]Damaged: get attackerTeam from projectile

sprunk:
Fixes the case where the attacker is captured/killed during
projectile flight which leaves the teamID incorrect/missing.


sprunk:
I thought it made sense to put that info there by default.

PR #394, opened 2018-08-19
fix MacOS glew.h include

MasterBel2:


rts/System/Platform/Mac/CrashHandler.cpp
```diff
@@ -20,12 +20,13 @@
 
 
 static bool HaveAddr2LineMac()
+{
 	static int i = -1;
 
 	if (i == -1) {
 		FILE* f = popen(ADDR2LINE " --help 2>/dev/null", "r");
 
-		if ((i = (f == nullptr)) == 1)
```
ashdnazg: Fix whitespace

rts/Rendering/Models/3DModel.h
```diff
@@ -259,10 +259,10 @@ struct S3DModel
 	void EnableAttribs() const;
 	void DisableAttribs() const;
 
-	void BindVertexArray() const { glBindVertexArray(vertexArray); }
+//    void BindVertexArray() const { glBindVertexArray(vertexArray); }
```
ashdnazg: uh, no

rtri:
All MacOS docs (and stackoverflow posts) I could find suggested that Apple GL headers live under `OpenGL/`, so any confirmation from some official source that it is actually just `GL/` would be nice.

The buildbot seems to be missing GLEW entirely and is a lost cause afaiac.

ashdnazg:
some projects I know moved entirely to https://glad.dav1d.de which will mean one dependency less

rtri:
No comment on the relative merits, but I think the path of least resistance would be to ship our own GLEW.

MasterBel2:
Haha, I think I understand how PRs work now, I made some really bad assumption when I first went through the tutorials trying to understand. Oops. Maybe it's best if this PR is just ignored.

That's sad about the buildbot :( My research agrees that default OpenGL headers live under `OpenGL/`, but glew.h isn't a default header? To build on mac, glew must be installed through macports/homebrew? Or… I mean, the two locations glew.h is on my machine is `/opt/local/include/GL/glew.h` (in the macports install directory) and `spring/include/GL/glew.h` – spring/ being the directory of the spring source? So pointing it to OpenGL/glew.h won't be successful, because that file categorically doesn't exist.

PR #395, opened 2018-08-26
Default uniforms & textures for map shaders

sanguinariojoe:
A way to ask spring to set the default uniforms and textures to Map LUA shaders, by means of a boolean optional parameter when calling ```SetMapShader```.

An example [here](https://github.com/sanguinariojoe/spring1944/blob/customMapShader/LuaUI/Widgets/gfx_customMapShader.lua), where both read shaders are just copies of the ones packaged within spring.

Some notes:

* I noticed some artifacts in Debug mode (with GLDEBUG)
* I created the PR just for the maintenance branch, because I cannot test development branch
* Related: https://springrts.com/mantis/view.php?id=6004

---

P.S. Sorry bacuase the initial PR commitment, I accidentally pressed enter...


rts/Lua/LuaOpenGL.cpp
```diff
@@ -4625,37 +4627,78 @@ int LuaOpenGL::GetWaterRendering(lua_State* L)
 
 int LuaOpenGL::GetMapRendering(lua_State* L)
 {
-	const char* key = luaL_checkstring(L, 1);
-
-	switch (hashString(key)) {
-		// float4
-		case hashString("splatTexScales"): {
```
abma: why was the switch/case replaced?

sanguinariojoe: It clearly trolled me when rebasing... I'm restoring case later!

sanguinariojoe:
I actually think that both, maintenance and develop, should have this feature. Unfortunately, I cannot test develop branch in this computer. So for the time being I'm creating just this PR to can discuss about it (let's see if other devs like this approach)

rtri:
I don't.

The *only* approach you should consider viable is creating a GetMapShaderUniform{Table} API so custom shaders can setup whatever engine defaults they need during the DrawGroundPre* callins.

sanguinariojoe:
@rtri, I created a similar PR with the `gl.GetMapShaderUniform`, but I need help to fix a LUA problem.

PR #396, opened 2018-08-27
gl.GetMapShaderUniform

sanguinariojoe:
Helper function, `gl.GetMapShaderUniform`, to get the required uniforms for `Spring.SetMapShader` set shaders. Also more options have been added to `gl.GetMapRendering`, to can conveniently set the flags for the map shaders.

An example [here](https://github.com/sanguinariojoe/spring1944/blob/customMapShader/LuaUI/Widgets/gfx_customMapShader.lua).

---

Obviously related with [that](https://github.com/spring/spring/pull/395#pullrequestreview-149537706).


rts/Lua/LuaOpenGL.cpp
```diff
@@ -4657,15 +4660,288 @@ int LuaOpenGL::GetMapRendering(lua_State* L)
...
 
+int LuaOpenGL::GetMapShaderUniform(lua_State* L)
+{
+	const int args = lua_gettop(L); // number of arguments
+	if ((args != 1) || !lua_istable(L, 1)) {
+		luaL_error(L, "Incorrect arguments to gl.GetMapShaderUniform(table)");
+	}
+
+	// Store the arguments in a list
+	std::vector<string> keys;
```
rtri: Use `std::array<const char*, 256> keys` and `unsigned int numUniforms` (clamped to `keys.size()`) here.

rts/Lua/LuaOpenGL.cpp
```diff
@@ -4657,15 +4660,288 @@ int LuaOpenGL::GetMapRendering(lua_State* L)
...
+			LOG_L(L_WARNING, "gl.GetMapShaderUniform: bad uniform name type");
+			return 0;
+		}
+		keys.push_back(lua_tostring(L, -1));
+	}
+
+	lua_createtable(L, keys.size(), 0);
+	int count = 1;
+	for (auto key : keys) {
+		if (key == "diffuseTex") {
```
rtri: Use `switch (hashString(key)) { ... }` here, as in `GetMapRendering`.

rtri:
> Do you know why?

Yes, the callins still have to be added to https://github.com/spring/spring/blob/develop/cont/LuaUI/callins.lua and https://github.com/spring/spring/blob/develop/cont/LuaUI/widgets.lua .

This is looking much better, thanks.

sanguinariojoe:
> Yes, the callins still have to be added to https://github.com/spring/spring/blob/develop/cont/LuaUI/callins.lua and https://github.com/spring/spring/blob/develop/cont/LuaUI/widgets.lua .

You nailed it!

sanguinariojoe:
I'm testing/fixing everything... When ready I'll let you know to eventually merge

rts/Lua/LuaOpenGL.cpp
sanguinariojoe: Done

rts/Lua/LuaOpenGL.cpp
sanguinariojoe: Done

sanguinariojoe:
Ok, after fixing some errors in the LUA stuff, it is working fine in my ATI GPU!

Tomorrow I'll try to test in the office NVidia, as well as to setup the develop branch version

rts/Lua/LuaOpenGL.cpp
```diff
@@ -4657,15 +4660,288 @@ int LuaOpenGL::GetMapRendering(lua_State* L)
...
+			return 1;
+		} break;
+		case hashString("parallaxMapping"): {
+			CSMFReadMap* smfMap = (CSMFReadMap*)readMap;
+			lua_pushboolean(L, smfMap->GetParallaxHeightTexture() != 0);
+			return 1;
+		} break;
+		// float
+		case hashString("baseDynamicMapLight"): {
+			CSMFReadMap* smfMap = (CSMFReadMap*)readMap;
```
rtri: Declare these variables once at the top, not in every case.

rts/Lua/LuaOpenGL.cpp
```diff
@@ -4657,15 +4660,288 @@ int LuaOpenGL::GetMapRendering(lua_State* L)
...
+			CSMFGroundDrawer* groundDrawer = (CSMFGroundDrawer*)smfMap->GetGroundDrawer();
+			const GL::LightHandler* lightHandler = groundDrawer->GetLightHandler();
+			lua_pushnumber(L, lightHandler->GetBaseLight());
+			return 1;
+		} break;
+		case hashString("maxDynamicMapLight"): {
```
rtri: `maxDynamicMapLights`

rts/Lua/LuaOpenGL.cpp
```diff
@@ -4744,217 +4744,222 @@ int LuaOpenGL::GetMapShaderUniform(lua_State* L)
...
+	std::array<char[64], 256> keys;
 	const int table = lua_gettop(L);
+	unsigned int numKeys = 0;
 	for (lua_pushnil(L); lua_next(L, table) != 0; lua_pop(L, 1)) {
 		if (!lua_israwstring(L, -1)) {
 			LOG_L(L_WARNING, "gl.GetMapShaderUniform: bad uniform name type");
 			return 0;
 		}
-		keys.push_back(lua_tostring(L, -1));
+		const char *s = lua_tostring(L, -1);
```
rtri: `strncpy(keys[numKeys], lua_tostring(L, -1), sizeof(keys[numKeys]));`
`keys[numKeys][63] = '\0';`
`numKeys = std::min(size_t(numKeys + 1), keys.size() - 1);`

sanguinariojoe:
It is working nice in my NVidia as well!

1 travis check has failed due to a checksum fail while downloading packages... Can you reboot/ignore it? Shall I do something? (Of course you can suggest further changes, so new commits force travis to run again)

I created a non-tested PR for the development version:
https://github.com/spring/spring/pull/397

rts/Lua/LuaOpenGL.cpp
```diff
@@ -4657,15 +4664,299 @@ int LuaOpenGL::GetMapRendering(lua_State* L)
...
+			} break;
+			case hashString("mapHeights"): {
+				lua_createtable(L, 2, 0);
+				lua_pushnumber(L, readMap->GetCurrMinHeight());
+				lua_rawseti(L, -2, 1);
+				lua_pushnumber(L, readMap->GetCurrMaxHeight());
+				lua_rawseti(L, -2, 2);
+			} break;
+			default: {
+				LOG_L(L_WARNING, "gl.Material: invalid map uniform '%s'", keys[count - 1]);
```
gajop: probably not gl.Material

rts/Lua/LuaOpenGL.cpp
```diff
@@ -4657,15 +4664,299 @@ int LuaOpenGL::GetMapRendering(lua_State* L)
...
+		strncpy(keys[numKeys], lua_tostring(L, -1), sizeof(keys[0]));
+		keys[numKeys++][sizeof(keys[0]) - 1] = '\0';
+		numKeys++;
+	}
+
+	lua_createtable(L, numKeys, 0);
+	for (unsigned int count = 0; count < numKeys; count++) {
+		switch (hashString(keys[count])) {
+			case hashString("diffuseTex"): {
+				lua_pushnumber(L, 0);
```
gajop: Constants..

rts/Lua/LuaOpenGL.cpp
```diff
@@ -4657,15 +4664,299 @@ int LuaOpenGL::GetMapRendering(lua_State* L)
...
+				lua_createtable(L, 3, 0);
+				for (int i = 0; i < 3; i++) {
+					lua_pushnumber(L, camera->GetPos()[i]);
+					lua_rawseti(L, -2, i + 1);
+				}
+			} break;
+			case hashString("groundAmbientColor"): {
+				lua_createtable(L, 3, 0);
+				for (int i = 0; i < 3; i++) {
+					lua_pushnumber(L, sunLighting->groundAmbientColor[i]);
```
gajop: functions for this already exist

rts/Lua/LuaOpenGL.cpp
```diff
@@ -4657,15 +4664,299 @@ int LuaOpenGL::GetMapRendering(lua_State* L)
...
+			case hashString("mapSizePO2"): {
+				lua_createtable(L, 2, 0);
+				lua_pushnumber(L, mapDims.pwr2mapx * SQUARE_SIZE * 1.0f);
+				lua_rawseti(L, -2, 1);
+				lua_pushnumber(L, mapDims.pwr2mapy * SQUARE_SIZE * 1.0f);
+				lua_rawseti(L, -2, 2);
+			} break;
+			case hashString("mapSize"): {
+				lua_createtable(L, 2, 0);
+				lua_pushnumber(L, mapDims.mapx * SQUARE_SIZE * 1.0f);
```
gajop: can be computed

sanguinariojoe:
> I don't understand why we need the GetMapShaderUniform function in the engine?
Most of those properties are already accessible: some are constants, some can be computed trivially, some are exposed in other functions.
I think we would be better served by having this helper function in the basecontent (Lua) archive (if anywhere), and only expose things which aren't already accessible.

I'm agree that getting duplicated stuff is not good at all. I'm also agree that moving this stuff to a LUA file would become cleaner, at first glance.

Said that... Collecting all the info right now is puzzling, slow and in some cases directly impossible. So I obviously want something centralised (either in the engine or in LUA).

IMO 3 main issues arise implementing that just in LUA:

1. Performance
2. Compatibility issues
3. Error hiding (More files to maintain, that are not compiled)

A sensible solution would be implementing some optimized LUA framework on top of this new function (in a similar fashion than unit custom shaders)

rts/Lua/LuaOpenGL.cpp
sanguinariojoe: Right!

rts/Lua/LuaOpenGL.cpp
```diff
@@ -4657,15 +4664,299 @@ int LuaOpenGL::GetMapRendering(lua_State* L)
 			lua_pushboolean(L, mapRendering->voidGround);
 			return 1;
 		} break;
+		case hashString("specularLighting"): {
```
gajop: Wrong name

rts/Lua/LuaOpenGL.cpp
```diff
@@ -4657,15 +4664,299 @@ int LuaOpenGL::GetMapRendering(lua_State* L)
...
+		case hashString("splatDetailNormalTexture"): {
+			lua_pushboolean(L, smfMap->GetSplatDistrTexture() != 0 &&
+			                   smfMap->HaveSplatNormalTexture());
+			return 1;
+		} break;
 		case hashString("splatDetailNormalDiffuseAlpha"): {
 			lua_pushboolean(L, mapRendering->splatDetailNormalDiffuseAlpha);
 			return 1;
 		} break;
+		case hashString("waterAbsortion"): {
```
gajop: This name also makes no sense. And btw, is voidWater not enough? (Maybe it's not, but can't tell from this code)

rts/Lua/LuaOpenGL.cpp
```diff
@@ -4657,15 +4664,299 @@ int LuaOpenGL::GetMapRendering(lua_State* L)
 			lua_pushboolean(L, mapRendering->voidGround);
 			return 1;
 		} break;
+		case hashString("specularLighting"): {
+			CSMFReadMap* smfMap = (CSMFReadMap*)readMap;
+			lua_pushboolean(L, smfMap->GetSpecularTexture() != 0);
+			return 1;
+		} break;
+		case hashString("splatDetailTexture"): {
```
gajop: Should be named correctly along with pretty much everything here.
This doesn't return a texture but a Boolean representing if the texture exists.

rts/Lua/LuaOpenGL.cpp
```diff
@@ -4657,15 +4664,299 @@ int LuaOpenGL::GetMapRendering(lua_State* L)
...
+		case hashString("haveShadows"): {
+			lua_pushboolean(L, shadowHandler.ShadowsLoaded());
+			return 1;
+		} break;
+		case hashString("haveInfoTex"): {
+			lua_pushboolean(L, infoTextureHandler->IsEnabled());
+			return 1;
+		} break;
+		// float
+		case hashString("baseDynamicMapLight"): {
```
gajop: Light information really doesn't belong here. 
It's even unclear if texture checks do - a more aptly named Spring.HasMapShadingTexture is a better idea.
See Spring.SetMapShadingTexture and similar existing functions

gajop:
> Said that... Collecting all the info right now is puzzling, slow and in some cases directly impossible. So I obviously want something centralised (either in the engine or in LUA).

This doesn't really make things clear, especially as you need to know the uniform names beforehand.

It's definitely not fast either. In fact if you call this function on each Draw call, you're probably writing really inefficient code (still something modern PCs wouldn't notice, but I didn't invoke the efficiency argument). Most uniforms you want to set only once, and then update a subset when things change (things like constants obviously won't change). In most cases only a few uniforms would change per Draw, and for that you'd do best keeping track by using specific callins (and modularized callouts)

But does this at least help with maintenance? Well sadly no, as someone still has to go and update these uniforms. Which they probably won't do if the default shader changes. There probably won't be any compilation bugs and no tests will fail. So it's not great for maintenance either. What would be great is a map shader framework (built-in to the engine base archive), with the engine using it by default (Lua shader by default, rather than optionally)

Imo, for this to be a good PR, it should separate stuff into a few modularized Spring.Getters, and maybe provide a few setters that are also missing.

sanguinariojoe:
OK, I'll make a 3rd PR deploying Getters for the missing data (and eventually some Setters)...

On top of that, we can discuss how to implement the LUA framework

gajop:
Sexy. I'll try to give you a hand once you have a working PR going

rtri:
gajop: generating too much noise.

1) a minor amount of duplication is acceptable, restricting it to the subset of hard-to-reach uniforms will just fragment the API with single-purpose getters
2) needing to know uniform names beforehand isn't a concern, the shader sources are available
3) the code could easily be refactored to declare all uniform names in one place so as to avoid the "things silently break if the default shader changes" scenario
4) you would not call this function every frame

> with the engine using (Lua shader by default, rather than optionally)

I don't think so.

gajop:
Maybe I was a bit too noisy indeed, sorry :) Not that this is a small post, but it feels necessary to write it.

> a minor amount of duplication is acceptable, restricting it to the subset of hard-to-reach uniforms will just fragment the API with single-purpose getters

I don't think it's bad to have an API with "single-purpose" getters. Although honestly I'd call them general-purpose as they don't have a very constrained use case (unlike a monolith "GetDefaultMapShaderUniforms" function)

> needing to know uniform names beforehand isn't a concern, the shader sources are available

It's a hassle to copy and update them in Lua. It can almost be guaranteed that games will forget to keep track of that. (the only solution I would recommend is an engine-maintained map shader framework)

> the code could easily be refactored to declare all uniform names in one place so as to avoid the "things silently break if the default shader changes" scenario

Sure, I concede some different code could do this differently.

> you would not call this function every frame

This is imo one of the more important issues to discuss and design for.
As I see it, there are a couple of ways one could go about this:
1. Query and update all (non-const) map uniforms each DrawMap frame regardless of what is updated. This is obviously the worst case solution.
2. Listen on existing events like `SunChanged` and query related values. This can be done using either existing getters or even this new function, but at that point the benefits of a single function diminish.
3. Create a new `MapUniformsChanged(uniform_names_table)` event that'd tell us exactly what changed. Would make things relatively simple to use, although quite likely not trivial to implement efficiently (reducing the number of Lua invocations).

Personally I prefer option 2 with existing (and as needed, new) Getters, since having a combination of Getters/Setters provides a clear interface on how to access and modify those properties to control the engine. Please tell me if you think this is a bad idea... I have been extending the engine access with various editable properties, and would like to know of a better design.

gajop:
Relevant changes were merged in https://github.com/spring/spring/pull/402 , closing

PR #398, opened 2018-08-31
Map shader (3rd way)

sanguinariojoe:
A 3rd minimalist way to can set all the default flags and uniforms to custom map shaders.

For the time being, the usage completely relies on the user LUA stuff. An example [here](https://github.com/sanguinariojoe/spring1944/blob/customMapShader/LuaUI/Widgets/gfx_customMapShader.lua).

---

Related with [this](https://github.com/spring/spring/pull/396)


gajop:
I don't know about the necessity of the code in `SetSquareTexGen`, but in case this is correct and complete (stuff works as far as you can see, right?), it can probably be immediately merged, regardless if you prefer my or kloot's way of doing things.

sanguinariojoe:
> I don't know about the necessity of the code in `SetSquareTexGen`

There is an assertion that says the shader should be bound, but custom LUA map shaders are always marked as unbound... `IProgramObject::Enable()` and `IProgramObject::Disable()` are just changing such 'bound' attribute

sanguinariojoe:
Would we add SunChanged to the callins?? (in a similar way to https://github.com/spring/spring/pull/398/commits/7e852ead1025b1791cf96a31c1efe22d54931ddb)

gajop:
Yes SunChanged should be added (basically all LuaUI related events from https://github.com/spring/spring/blob/develop/rts/System/Events.def should exist).

Listening to those kind of events would allow you to update only the relevant uniforms. Best start with a working version and then move to a more optimized (and modular one). Especially since I can imagine custom map shaders wanting modular components (e.g. camera and sun, but maybe not all of the textures)

sanguinariojoe:
OK, the example works now like a charm!

I'm creating the SunChanged callins, updating the example, and coming back to you

sanguinariojoe:
Here we go! The [example](https://github.com/sanguinariojoe/spring1944/blob/customMapShader/LuaUI/Widgets/gfx_customMapShader.lua) is optimized! You can test the `SunChanged` callin adding this code at the end of such script:

```
function widget:GameFrame(frame)
    local t = frame / 30
    local x, y, z = glGetSun("pos")
    x = math.sin(2 * math.pi * t / 10.0)
    local l = math.sqrt(x * x + y * y + z * z)
    Spring.SetSunDirection(x / l, y / l, z / l)
end
```

sanguinariojoe:
Jajaja, I did not pushed the changes! Now you can visit the [example](https://github.com/sanguinariojoe/spring1944/blob/customMapShader/LuaUI/Widgets/gfx_customMapShader.lua).

PR #401, opened 2018-09-16
API for changing texture filtering params. Not tested yet.

lhog:
Let's see if it even compiles on the server.

### DIFF
--- a/rts/Lua/LuaOpenGL.cpp
+++ b/rts/Lua/LuaOpenGL.cpp
@@ -243,6 +243,7 @@ bool LuaOpenGL::PushEntries(lua_State* L)
 
 	REGISTER_LUA_CFUNC(Texture);
 	REGISTER_LUA_CFUNC(CreateTexture);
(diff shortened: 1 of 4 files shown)


rts/Lua/LuaOpenGL.cpp
```diff
@@ -2728,6 +2729,49 @@ int LuaOpenGL::CreateTexture(lua_State* L)
...
+{
+	if (!lua_isstring(L, 1))
+		return 0;
+
+	const std::string& texName = luaL_checkstring(L, 1);
+
+	const LuaTextures& textures = CLuaHandle::GetActiveTextures(L);
+	const LuaTextures::Texture* tex = textures.GetInfo(texName);	
+
+	//LuaTextures::Texture newTex = *tex;
```
rtri: This needs a non-const accessor in LuaTextures s.t. `tex->target` (and other members) can be changed directly, othewise any unspecified parameters will revert to their defaults.

lhog:
Cool thank you so much. I was about to test if my implementation works.

PR #405, opened 2018-09-30
Make large portions of CMake code lowercase

gajop:
Our code has a lot of inconsistencies with CMake code style, making it harder to maintain or even just read. 

I figured the first step in the cleanup is to at least get it to conform to a specific case. We are currently using CAPITAL_CASE, Capitalized_Case, and lower_case, sometimes even all together in the same file. This converts most commands to lowercase (easiest to type, so we might as well pick that style). Other than that, this PR cleans up whitespace and comments in a few places, but generally doesn't try to do anything smart.

There's going to be some cleanup after that, but we need to start somewhere.


rtri:
+1 from me because I despise This_code_Style.

Anyone against?

ashdnazg:
+9000 from me

PR #407, opened 2018-10-02
Lua defs: movedef.speedModClass accepts string

sprunk:
Same rules as the current name-based fallback.

This is because currently `speedModClass` only accepts numbers but these are magic and not exposed to Lua.


sprunk:
`Game` is not exposed to defs parsing (see #370) so the new `Game.speedModClasses` table isn't very useful at that point and magic numbers still have to be used.

PR #410, opened 2018-10-17
not needed by macos buildbot [Question]

TurBoss:
why is include specified to be OPENGL if __APPLE__?

none of the distributions of glew (brew or macSDK) install it in that path

instead they use GL/glew.h

Thanks

### DIFF
--- a/buildbot/slave/osx/macos_sierra_x64.cmake
+++ b/buildbot/slave/osx/macos_sierra_x64.cmake
@@ -9,8 +9,6 @@ set(Boost_USE_STATIC_LIBS  TRUE CACHE BOOL "")
 
 set(PRD_JSONCPP_INTERNAL FALSE CACHE BOOL "")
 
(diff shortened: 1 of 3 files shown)


gajop:
I remember this discussion (but not its source), when we went the opposite direction: https://github.com/spring/spring/commit/95970341b26cbb4e4e37ec570cb5bdb1dda48abc#diff-bd7aba100609d85a472ac846bd7714de 
Is there any verified reference on what the proper path is?

TurBoss:
I know 3 options

1. Adding the full path to glew include dir in cmake

glew from macosSDK

 ```
set(GLEW_INCLUDE_DIR "/usr/local/include/GL" CACHE STRING "")
#include <glew.h>
```
or glew from home-brew
```
set(GLEW_INCLUDE_DIR "/usr/local/Cellar/glew/2.1.0/include/GL/" CACHE STRING "")
#include <glew.h>
 ```

2. Don't' adding glew include dir

```
#include <GL/glew.h>
```
or
```
#include <OPENGL/glew.h>
```

PR #413, opened 2018-10-24
default to core 4 context

abma:
any complains against this?

this makes spring to run again with default settings on linux with ATI-OS-Drivers.


rtri:
Spring is *not* ready to go core yet, it will run but only with (highly predictable, already known) errors.

As for "ATI-OS", the latest mesa/radeonsi release supports up to 4.4 compat. I can safely guesstimate there are less than five people who 1) use a different ATI driver and 2) have any interest in testing current develop, so making it "work" out of the box has limited value.

abma:
ok, thanks for the info :)

PR #415, opened 2018-11-04
WIP: clang-tidy checks

gajop:
Lately I've been working to get clang-tidy running for Spring, and doing some cleanup in the process.

[Clang-tidy](http://clang.llvm.org/extra/clang-tidy/)  is a useful tool for modernizing code, performing micro optimizations and even finding bugs (provided you use it carefully and don't introduce them instead!)

Right now I've setup clang-tidy to run while Spring is being built, and I've made it do only the checks that I wanted to deal with initially - as we go long the list of checks should probably be increased, but I think the first priority would be to get the warning count to 0 (by fixing diagnostics or explicitly [suppressing them](http://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics)).Unless we change our coding style, some checks will probably be disabled forever, e.g. `-readability-braces-around-statements`.

Performing clang-tidy while compiling can be awfully slow (didn't measure but build times seemed to have tripled at least) which feels largely unnecessary for the first build. 

During development I would like to have a way to turn it on for builds, e.g. my workflow would likely go as  follows:
1. build without clang-tidy
2. do some changes
3. build with clang-tidy

I would also like to do some refactoring from time to time:
1. perform full builds with clang-tidy
2. do some refactoring/global cleanup

And it would be nice to have it run on buildbots:
1. perform full builds with clang-tidy

CMake is not my strong suite, so if anyone's willing to lend a hand I'd be grateful.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,31 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.6)
 
+find_program(
+  CLANG_TIDY_EXE
+  NAMES "clang-tidy"
+  DOC "Path to clang-tidy executable"
+)
+if(NOT CLANG_TIDY_EXE)
+  message(STATUS "clang-tidy not found.")
+else()
+  message(STATUS "clang-tidy found: ${CLANG_TIDY_EXE}")
+  string(CONCAT CLANG_TIDY_CHECKS
+  	# disable all by default
+  	"-*"
+	# enable a few general categories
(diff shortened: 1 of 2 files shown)


abma:
why not just add a clang-tidy config file? i don't see why this should be added to the cmake files.

gajop:
> why not just add a clang-tidy config file? i don't see why this should be added to the cmake files.

You are right, it should just be the config file & scripts to execute it. I know this now =)

PR #416, opened 2018-11-08
Cleanup

gajop:
A few issues are unresolved (left in FIXMEs, hoping for some comment before I tackle it), but other than that this gets me to 0 clang-tidy warnings with my current (uncommitted) check setup, yey.

Some fixes might not be super obvious (like reorganizing constructor parameters in `GameServer.cpp` to satisfy `-Wreorder`), so feel free to open up a discussion.

### DIFF
--- a/rts/Game/UI/QuitBox.cpp
+++ b/rts/Game/UI/QuitBox.cpp
@@ -383,7 +383,7 @@ void CQuitBox::MouseMove(int x, int y, int dx, int dy, int button)
 		const float tsz = sz / float(MAX_QUIT_TEAMS);
 
 		// ??
-		*(volatile int*) &startTeam = std::max(0, std::min((int)lround(scr / tsz), MAX_QUIT_TEAMS - numTeamsDisp));
+		*(volatile int*) &startTeam = std::max(0, std::min((int)std::lround(scr / tsz), MAX_QUIT_TEAMS - numTeamsDisp));
 		return;
 	}
 
--- a/rts/Game/UI/QuitBox.cpp
+++ b/rts/Game/UI/QuitBox.cpp
@@ -2,6 +2,8 @@
 
 #include "QuitBox.h"
 
+#include <cmath>
+
 #include "MouseHandler.h"
 #include "Game/GameSetup.h"
 #include "Game/GlobalUnsynced.h"
@@ -21,7 +23,6 @@
 #include "System/MsgStrings.h"
 
 #include <SDL_keycode.h>
-
 
 #define MAX_QUIT_TEAMS (teamHandler.ActiveTeams() - 1)
 
@@ -382,7 +383,7 @@ void CQuitBox::MouseMove(int x, int y, int dx, int dy, int button)
 		const float tsz = sz / float(MAX_QUIT_TEAMS);
 
 		// ??
-		*(volatile int*) &startTeam = std::max(0, std::min((int)(scr / tsz + 0.5), MAX_QUIT_TEAMS - numTeamsDisp));
+		*(volatile int*) &startTeam = std::max(0, std::min((int)lround(scr / tsz), MAX_QUIT_TEAMS - numTeamsDisp));
 		return;
 	}
 
--- a/rts/Game/UI/ShareBox.cpp
+++ b/rts/Game/UI/ShareBox.cpp
@@ -389,7 +389,7 @@ void CShareBox::MouseMove(int x, int y, int dx, int dy, int button)
 		float tsz = sz / float(MAX_SHARE_TEAMS);
 
 		// ??
-		*(volatile int*) &startTeam = std::max(0, std::min((int)lround(scr / tsz), MAX_SHARE_TEAMS - numTeamsDisp));
+		*(volatile int*) &startTeam = std::max(0, std::min((int)std::lround(scr / tsz), MAX_SHARE_TEAMS - numTeamsDisp));
 		return;
 	}
 
--- a/rts/Game/UI/ShareBox.cpp
+++ b/rts/Game/UI/ShareBox.cpp
@@ -2,6 +2,8 @@
 
 #include "ShareBox.h"
 
+#include <cmath>
+
 #include "MouseHandler.h"
 #include "Game/GlobalUnsynced.h"
 #include "Game/SelectedUnitsHandler.h"
@@ -16,7 +18,6 @@
 #include "System/MsgStrings.h"
 
 #include <SDL_keycode.h>
-
 
 #define MAX_SHARE_TEAMS (teamHandler.ActiveTeams() - 1)
 int CShareBox::lastShareTeam = 0;
@@ -388,7 +389,7 @@ void CShareBox::MouseMove(int x, int y, int dx, int dy, int button)
 		float tsz = sz / float(MAX_SHARE_TEAMS);
 
 		// ??
-		*(volatile int*) &startTeam = std::max(0, std::min((int)(scr / tsz + 0.5), MAX_SHARE_TEAMS - numTeamsDisp));
+		*(volatile int*) &startTeam = std::max(0, std::min((int)lround(scr / tsz), MAX_SHARE_TEAMS - numTeamsDisp));
 		return;
 	}
 
--- a/rts/Lua/LuaUnitDefs.cpp
+++ b/rts/Lua/LuaUnitDefs.cpp
@@ -75,8 +75,9 @@ bool LuaUnitDefs::PushEntries(lua_State* L)
 	const std::array<const IndxFuncType, 3> indxFuncs = {{UnitDefIndex, UnitDefNewIndex, UnitDefMetatable}};
 	const std::array<const IterFuncType, 2> iterFuncs = {{Pairs, Next}};
 
-	for (auto it = defsVec.cbegin(); it != defsVec.cend(); ++it) {
-		const auto def = unitDefHandler->GetUnitDefByID(it->id);
+	for (const auto& unitDef: defsVec) {
+		// FIXME: This makes no sense, we already have the def, it's valid.
+		const auto def = unitDefHandler->GetUnitDefByID(unitDef.id);
 
 		if (def == nullptr)
 			continue;
--- a/rts/Lua/LuaUnitDefs.cpp
+++ b/rts/Lua/LuaUnitDefs.cpp
@@ -76,7 +76,7 @@ bool LuaUnitDefs::PushEntries(lua_State* L)
 	const std::array<const IterFuncType, 2> iterFuncs = {{Pairs, Next}};
 
 	for (const auto& unitDef: defsVec) {
-		// FIXME: This makes no sense, we already have the def, it's valid.
+		// The first unitDef is invalid (dummy), so we skip it
 		const auto def = unitDefHandler->GetUnitDefByID(unitDef.id);
 
 		if (def == nullptr)
--- a/rts/Lua/LuaWeaponDefs.cpp
+++ b/rts/Lua/LuaWeaponDefs.cpp
@@ -74,8 +74,9 @@ bool LuaWeaponDefs::PushEntries(lua_State* L)
 	const std::array<const IndxFuncType, 3> indxFuncs = {{WeaponDefIndex, WeaponDefNewIndex, WeaponDefMetatable}};
 	const std::array<const IterFuncType, 2> iterFuncs = {{Pairs, Next}};
 
-	for (auto it = defsVec.cbegin(); it != defsVec.cend(); ++it) {
-		const auto def = weaponDefHandler->GetWeaponDefByID(it->id);
+	for (const auto& weaponDef: defsVec) {
+		// FIXME: This makes no sense, we already have the def, it's valid.
+		const auto def = weaponDefHandler->GetWeaponDefByID(weaponDef.id);
 
 		if (def == nullptr)
 			continue;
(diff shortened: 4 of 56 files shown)


rts/System/Log/DefaultFilter.cpp
```diff
@@ -22,7 +22,7 @@
 extern "C" {
 #endif
 
-
+// NOLINTNEXTLINE{readability-redundant-declaration}
 bool log_frontend_isEnabled(int level, const char* section);
```
rtri: this can go

rts/System/myMath.cpp
```diff
@@ -28,7 +28,10 @@ void CMyMath::Init()
 	}
 
 	unsigned checksum = 0;
+	// NOLINTNEXTLINE{modernize-loop-construct}
```
rtri: 1) the number of NOLINT's is already getting worrysome, source code should not be contorted to satisfy any single tool imo.

2) if float* and int* didn't have the same alignment, this code would have crashed and burned a long time ago.

rts/System/Platform/Watchdog.cpp
```diff
@@ -249,6 +249,7 @@ namespace Watchdog
 		}
 
 		if (0 == --(threadInfo->numreg))
+			// FIXME: undefined behavior, destination object type 'Watchdog::WatchDogThreadInfo' is not TriviallyCopyable [bugprone-undefined-memory-manipulation]
```
rtri: `*threadInfo = {};`

rts/Net/NetCommands.cpp
```diff
@@ -226,7 +226,7 @@ float CGame::GetNetMessageProcessingTimeLimit() const
 	return Clamp(simDrawRatio * gu->avgSimFrameTime, 5.0f, 1000.0f / gu->minFPS);
 }
 
-void CGame::ClientReadNet()
+void CGame::ClientReadNet() // NOLINT{readability-function-size}
```
rtri: I wonder what the threshold for this rule is...

rts/Game/UI/QuitBox.cpp
```diff
@@ -382,7 +383,7 @@ void CQuitBox::MouseMove(int x, int y, int dx, int dy, int button)
 		const float tsz = sz / float(MAX_QUIT_TEAMS);
 
 		// ??
-		*(volatile int*) &startTeam = std::max(0, std::min((int)(scr / tsz + 0.5), MAX_QUIT_TEAMS - numTeamsDisp));
+		*(volatile int*) &startTeam = std::max(0, std::min((int)lround(scr / tsz), MAX_QUIT_TEAMS - numTeamsDisp));
```
rtri: `std::`

rts/Game/UI/ShareBox.cpp
```diff
@@ -388,7 +389,7 @@ void CShareBox::MouseMove(int x, int y, int dx, int dy, int button)
 		float tsz = sz / float(MAX_SHARE_TEAMS);
 
 		// ??
-		*(volatile int*) &startTeam = std::max(0, std::min((int)(scr / tsz + 0.5), MAX_SHARE_TEAMS - numTeamsDisp));
+		*(volatile int*) &startTeam = std::max(0, std::min((int)lround(scr / tsz), MAX_SHARE_TEAMS - numTeamsDisp));
```
rtri: `std::`

rts/Lua/LuaWeaponDefs.cpp
```diff
@@ -370,9 +371,9 @@ static int CustomParamsTable(lua_State* L, const void* data)
 	const spring::unordered_map<std::string, std::string>& params = *((const spring::unordered_map<std::string, std::string>*)data);
 	lua_newtable(L);
```
rtri: unrelated, but as good a time as any: `lua_createtable(L, 0, params.size());`

rts/Rendering/Env/Decals/GroundDecalHandler.cpp
```diff
@@ -769,8 +769,10 @@ void CGroundDecalHandler::TestScarOverlaps(const Scar& scar)
 
 	for (int y = y1; y <= y2; ++y) {
 		for (int x = x1; x <= x2; ++x) {
-			auto& quad = scarField[y * scarFieldX+ x];
+			auto& quad = scarField[y * scarFieldX + x];
 
+			// Can't tell if quad will change in below loop, so:
```
rtri: it won't

rts/Rendering/Env/Particles/Classes/ShieldSegmentProjectile.cpp
```diff
@@ -43,7 +43,7 @@ static spring::unsynced_map<const AtlasedTexture*, std::vector<float2> > spheret
 
 
 
-ShieldSegmentCollection& ShieldSegmentCollection::operator = (ShieldSegmentCollection&& ssc) {
+ShieldSegmentCollection& ShieldSegmentCollection::operator=(ShieldSegmentCollection&& ssc) noexcept {
```
rtri: a "spaces versus no spaces for overloaded operators" rule was never established, this could be a long road

rts/Rendering/HUDDrawer.cpp
```diff
@@ -172,7 +173,7 @@ void HUDDrawer::DrawWeaponStates(const CUnit* unit)
 		}
 		if (w->reloadStatus > gs->frameNum) {
 			font->SetTextColor(0.8f, 0.2f, 0.2f, 0.8f);
-			font->glFormat(-0.9f * 0.5f + 0.5f, yPos * 0.5f + 0.5f, fontSize, FONT_SCALE | FONT_NORM | FONT_BUFFERED, "%s: Reloading (%i%%)", wd->description.c_str(), 100 - int(100.0f * (w->reloadStatus - gs->frameNum) / int(w->reloadTime / unit->reloadSpeed) + 0.5f));
+			font->glFormat(-0.9f * 0.5f + 0.5f, yPos * 0.5f + 0.5f, fontSize, FONT_SCALE | FONT_NORM | FONT_BUFFERED, "%s: Reloading (%i%%)", wd->description.c_str(), 100 - roundf(100.0f * (w->reloadStatus - gs->frameNum) / int(w->reloadTime / unit->reloadSpeed)));
```
rtri: `std::`

rts/Rendering/HUDDrawer.cpp
```diff
@@ -162,7 +163,7 @@ void HUDDrawer::DrawWeaponStates(const CUnit* unit)
 		if (wd->stockpile && !w->numStockpiled) {
 			if (w->numStockpileQued > 0) {
 				font->SetTextColor(0.8f, 0.2f, 0.2f, 0.8f);
-				font->glFormat(-0.9f * 0.5f + 0.5f, yPos * 0.5f + 0.5f, fontSize, FONT_SCALE | FONT_NORM | FONT_BUFFERED, "%s: Stockpiling (%i%%)", wd->description.c_str(), int(100.0f * w->buildPercent + 0.5f));
+				font->glFormat(-0.9f * 0.5f + 0.5f, yPos * 0.5f + 0.5f, fontSize, FONT_SCALE | FONT_NORM | FONT_BUFFERED, "%s: Stockpiling (%i%%)", wd->description.c_str(), roundf(100.0f * w->buildPercent));
```
rtri: `std::`

rts/Rendering/Models/3DModel.cpp
```diff
@@ -599,8 +599,8 @@ void LocalModel::UpdateBoundingVolume()
 				float3(pMins.x,  pMaxs.y,  pMaxs.z),
 			};
 
-			for (unsigned int k = 0; k < 8; k++) {
-				const float3 vertex = matrix * verts[k];
+			for (const auto& v: verts) {
```
rtri: `const float3& v:`

rts/Rendering/GL/glExtra.cpp
```diff
@@ -517,8 +517,8 @@ void gleGenMeshBuffers(unsigned int* meshData) {
 	}
 
 	{
-		for (size_t n = 0; n < CYL_INDCS.size(); n++) { CYL_INDCS[n] += (BOX_VERTS.size()                   ); }
-		for (size_t n = 0; n < SPH_INDCS.size(); n++) { SPH_INDCS[n] += (BOX_VERTS.size() + CYL_VERTS.size()); }
+		for (auto& cyl_indc: CYL_INDCS) cyl_indc += BOX_VERTS.size();
```
rtri: this might be an example where ranged loops express intent less clearly.

rts/System/myMath.cpp
gajop: Wall of text but: 
1. The number of NOLINTs is a bit large, although most of them are trying to silence dead code `false && expr` or `true || expr`, and will probably go away once that section is fixed/implemented. 
However I agree with you on principle, and I'll try to get rid as many as I can by disabling some style checks globally, but I'd like to leave a few of them in place. I suppose it can serve as an extra `don't change this code` warning, but more importantly, it allows us to use clang-tidy with a 0 warnings setup. 
As long as it doesn't produce too many false positives, any tool that helps reduce the errors, increases performance and keeps code readable is important to use in my opinion.
In addition, I think C++ as a language is taking the path of annotating the code more to ensure safety, readability and to support tools issuing warning/error checks (https://en.cppreference.com/w/cpp/language/attributes#Standard_attributes). 

2. To the best of my understanding, this is UB that happens to be correct because of our compilers/target systems and maybe compile flags(not sure if we're using `-fno-strict-aliasing`). It seems to be a violation of strict aliasing rules: [[1]](https://stackoverflow.com/questions/98340/what-are-the-common-undefined-unspecified-behavior-for-c-that-you-run-into) [[2]](https://cellperformance.beyond3d.com/articles/2006/06/understanding-strict-aliasing.html)

rts/Rendering/Env/Particles/Classes/ShieldSegmentProjectile.cpp
gajop: Hm, agreed, code seems mixed on that. Think I'll avoid doing any (extra) changes.

rts/Rendering/GL/glExtra.cpp
gajop: I thought the same while I was writing it, so maybe `// NOLINT` it?

Then again it might just be something that gets clear once you get used to it. `For each element in $X: add to it`is almost like how'd you say it in English with a different word order.

rts/Lua/LuaUnitDefs.cpp
```diff
@@ -75,8 +75,9 @@ bool LuaUnitDefs::PushEntries(lua_State* L)
 	const std::array<const IndxFuncType, 3> indxFuncs = {{UnitDefIndex, UnitDefNewIndex, UnitDefMetatable}};
 	const std::array<const IterFuncType, 2> iterFuncs = {{Pairs, Next}};
 
-	for (auto it = defsVec.cbegin(); it != defsVec.cend(); ++it) {
-		const auto def = unitDefHandler->GetUnitDefByID(it->id);
+	for (const auto& unitDef: defsVec) {
+		// FIXME: This makes no sense, we already have the def, it's valid.
```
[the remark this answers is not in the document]
gajop: Ahh I see. I misread the `IsValidUnitDefID` method. I'll rewrite the comment but I won't do any larger change here. 
But yea, I agree we start counting at one or maybe change the approach altogether.

rts/System/Platform/Watchdog.cpp
gajop: Oh cool, was this an actual UB?

gajop:
Thanks for the review as usual! I'll wait for your response and then do any additional fixes as necessary.

rts/System/Platform/Watchdog.cpp
rtri: memset'ing a std::shared_ptr wasn't kosher, although the control flow prevented any real damage from being done.

rts/Map/SMF/SMFReadMap.cpp
```diff
@@ -61,6 +61,7 @@ CSMFReadMap::CSMFReadMap(const std::string& mapName): CEventClient("[CSMFReadMap
 	haveSplatDetailDistribTexture = (!mapInfo->smf.splatDetailTexName.empty() && !mapInfo->smf.splatDistrTexName.empty());
 	haveSplatNormalDistribTexture = false;
 
+	// FIXME: undefined behavior, destination object type 'MapTexture' is not TriviallyCopyable [bugprone-undefined-memory-manipulation]
```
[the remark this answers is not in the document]
rtri: technically UB then, but uninteresting because this LOC runs before the textures are assigned.

since MapTexture zero-inits itself anyway the memset is actually redundant.

rts/System/myMath.cpp
gajop: Those NOLINTs can go away if you want me to ignore [readability-simplify-boolean-expr](https://clang.llvm.org/extra/clang-tidy/checks/readability-simplify-boolean-expr.html) and [misc-static-assert](https://clang.llvm.org/extra/clang-tidy/checks/misc-static-assert.html) checks globally. Honestly I'd let them stay as those blocks of code *should* alert the user that something is odd (the actual check is disabled). Up to you.
PS: I'm thinking of replacing NOLINT(specific-check) with just NOLINT. It seems the whole // NOLINT  thing is used by other tools as well (e.g. [cpplint](https://en.wikipedia.org/wiki/Cpplint))

rts/Map/SMF/SMFReadMap.cpp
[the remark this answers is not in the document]
gajop: No bugs found, but oh well :)  At least it's not a false positive

rts/Rendering/GL/glExtra.cpp
gajop: fwiw clang-tidy isn't producing any more warnings related to the checks I have enabled, so it's not like I would be spreading NOLINTs to more places (unless we decide to turn some other checks on). 
That said, yea, I'd prefer to use them to replace iterator based loops instead.

rts/System/myMath.cpp
rtri: Ignore them for now, I'll see which of these instances (probably most) can be removed or converted to comments later.

gajop:
Anything else left (for me to do)? Is it ready for merge?

rtri:
Nope, thanks.

PR #419, opened 2018-11-10
fix #5537

gajop:
ArchiveScanner will now properly detect when modinfo.lua/mapinfo.lua
 changes, and reparse the archive.
This fixes issues like changing the depends subtable and not seeing any
 change until the cache dir is deleted.

Implementation details:
- We now cache an additional path to the modinfo.lua/mapinfo.lua and
 the related modified timestamp. This applies to datadirs (.sdd)
 archives only
- This means that reading the cache will have two additional stat calls
 for datadirs. I have over 40 .sdd archives (on both HDD and SSD) and
 there is no noticeable performance hit. Players are likely to have very
 few of such archives, probably 0. So unlikely to hurt performance in
 general.
- Likewise, this information is only saved for datadirs as well, so
there shouldn't be any performance hit on writing the cache either.
- The way I obtain the full path of the archive is stolen from
`CVFSHandler::GetFileAbsolutePath`, and maybe could be rewritten with
DRY. Also probably should be tested on windows, the "/" seems suspicious

Minor:
- INTERNAL_VAR is incremented
- CheckCachedData signature is modified to pass modifiedTime by
reference as it cannot be nullptr
- internalver is loaded as lowercase (consistent to how its saved)

### DIFF
--- a/rts/System/FileSystem/ArchiveScanner.cpp
+++ b/rts/System/FileSystem/ArchiveScanner.cpp
@@ -13,6 +13,7 @@
 #include "ArchiveLoader.h"
 #include "DataDirLocater.h"
 #include "Archives/IArchive.h"
+#include "Archives/DirArchive.h"
 #include "FileFilter.h"
 #include "DataDirsAccess.h"
 #include "FileSystem.h"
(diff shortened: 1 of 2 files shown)


gajop:
The diff in CheckCachedData is a bit unfortunately displayed. I just did an early return as I strongly dislike code like this:
```
if (x) {
// many lines of code
}
return false;
```

PR #422, opened 2018-11-30
SAT based collision detection (develop branch)

sanguinariojoe:
A collision detector based on Separable Axes Theorem (SAT). It is a bit more expensive than the radial check, but make possible to model outstretched bodies:

https://youtu.be/L2W0SO9i62w

Comments:
* WARNING! Very little tested!
* Since the model is more precise, I would say is better to take the unitDef footprint values, and in case they are not available, the moveDef ones. That way, default values can be set by means of the move type, and per-unit overridden. This should be backward compatible, provided that in documentation it's stated that both values shall match. 
* The repulsion model (when a collision is detected) is not affected. But in my tests the effect of the approximations seems negligible.
* The y-direction (bottom-up) interactions are not affected at all, i.e. in that case the radial chek is still used to determine if 2 solids are colliding.

### DIFF
--- a/rts/ExternalAI/SSkirmishAICallbackImpl.cpp
+++ b/rts/ExternalAI/SSkirmishAICallbackImpl.cpp
@@ -1443,10 +1443,8 @@ EXPORT(bool) skirmishAiCallback_Cheats_isEnabled(int skirmishAIId) {
 
 EXPORT(bool) skirmishAiCallback_Cheats_setEnabled(int skirmishAIId, bool enabled)
 {
-	AI_CHEAT_FLAGS[skirmishAIId].first = enabled;
-
-	if (enabled && !AI_CHEAT_FLAGS[skirmishAIId].second) {
-		LOG("SkirmishAI (ID = %i, team ID = %i) is using cheats!", skirmishAIId, AI_TEAM_IDS[skirmishAIId]);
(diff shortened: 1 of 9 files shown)


ashdnazg:
maintenance should only get commits by cherry-picking, so we should only work on this PR.
If this is tested to be ok, I suspect caching {x,z}size * 0.5f * SQUARE_SIZE for each unit can save some computations, but that might need profiling.
Also it's worth noting that if a unit is above another unit, you will use the radius rather than height. (this might be ok, just writing it here so it doesn't surprise anyone).

sanguinariojoe:
> maintenance should only get commits by cherry-picking, so we should only work on this PR.

I'm agree!

> I suspect caching {x,z}size * 0.5f * SQUARE_SIZE for each unit can save some computations

Related to that, I'm not sure if I messed up with x/z directions... It is working with out spring 1944 GERMal model...

> Also it's worth noting that if a unit is above another unit, you will use the radius rather than height. (this might be ok, just writing it here so it doesn't surprise anyone).

Thanks! I updated the description

rtri:
The SA test should probably not be enabled unless at least one party wants the extra precision. You can implement this as a boolean flag for now, set to true when an object's footprint is non-square and a global modrule allows it.

I'll have to think about letting one type of *Def override values from another.

sanguinariojoe:
Sorry, I messed up with the branches... I'm reopening it again soon to continue discussing

PR #425, opened 2018-12-03
SAT based collision detection (Don't merge yet)

sanguinariojoe:
_Restored PR. See https://github.com/spring/spring/pull/422_

A collision detector based on Separable Axes Theorem (SAT). It is a bit more expensive than the radial check, but make possible to model outstretched bodies:

https://youtu.be/L2W0SO9i62w

Comments:

- WARNING! Very little tested! Probably some sort of profiling test is worthing
- ~~Now it is possible to define default footprint values at the moveDefs, optionally overriding them later in the unitDefs. This should be backward compatible, provided that in documentation it's stated that both values shall match. **Should we enforce odd values?**~~
- The repulsion model (when a collision is detected) is not affected yet.
- **Right now is possible to get units stucked**. I'm working on that, but **hints are welcome**.
- The y-direction (bottom-up) interactions are not affected at all, i.e. in that case the radial check is still used to determine if 2 solids are colliding.
- A new game/mod option is available to determine whether SAT-based collision detector should be considered or not
- SAT-based collision detector is not applied if both colliders have square footprint shapes

### DIFF
--- a/rts/Sim/Misc/ModInfo.cpp
+++ b/rts/Sim/Misc/ModInfo.cpp
@@ -31,6 +31,7 @@ void CModInfo::ResetState()
 	allowUnitCollisionOverlap = true;
 	allowGroundUnitGravity    = true;
 	allowHoverUnitStrafing    = true;
+	useSATCollisionDetection  = false;
 
 	constructionDecay      = true;
 	constructionDecayTime  = 1000;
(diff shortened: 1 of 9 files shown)


sanguinariojoe:
OK, @rtri was right that trying to use unitDef footprints is hard, due to the path finding (which relies just on moveDefs). It would be very inefficient otherwise...

I am still trying to fix the eventually blocked units. Again, hints are welcome!

rtri:
I have no hints to offer, occasional stuck units are a fact of life. There was however a typo in the `checkSAT` math that perhaps contributed.

Merged (with gracias) because I don't want to delay this indefinitely, serious problems can always be removed in followup PR's.

PR #426, opened 2018-12-08
attempt to fix some build steps on macos buildbot

TurBoss:
this uses the system linker from apple instead of the gnu ones that fixes some build steps


TurBoss:
Thanks

PR #432, opened 2019-02-16
Configurable drag selection and area command leeway v3

GoogleFrog:
This PR adds the following springsettings:
- MouseDragSelectionThreshold
- MouseDragCircleCommandThreshold
- MouseDragBoxCommandThreshold
- MouseDragFrontCommandThreshold

Godde keeps commenting that he often right clicks on a unit with the intention to target it, but while doing so accidentally moves the mouse very slightly (because he is microing so hard). This movement leads the interface to interpret his input as an area command, which is likely to not include the desired unit. I have tried hacking some solutions for special cases, but it is a bit of a waste of time when the thresholds are so easily changed within the engine.

See http://zero-k.info/Forum/Post/197424#197424 and https://github.com/ZeroK-RTS/Zero-K/issues/3025

I also fixed an area command visualisation bug caused by mismatched inequalities.


ashdnazg:
ty

PR #434, opened 2019-03-04
Update CircuitAI {1.0.4}

rlcevg:
Please merge into maintenance.

Btw, is it better to create 2 separate pull requests (develop + maintenance) from my side, or only PR for develop (and ask to populate it into maintenance)?


ashdnazg:
doesn't matter much I think

PR #435, opened 2019-03-13
Fix #6158 by correctly using the side effect of AutoTarget().

GoogleFrog:


GoogleFrog:
I screwed up the whitespace, so it is probably better to make the change manually than pull the PR.

rtri:
See mantis, doesn't solve the issue.

PR #437, opened 2019-03-20
Add `GetUnitCurrentCommand` lua callout

sprunk:
Spring.GetUnitCurrentCommand(unitID) -> [cmdID, cmdOpts (coded), cmdTag[, cmdParam1[, cmdParam2 ... ]]]

Options are coded as a single number and parameters are returned on the stack so that there are no table allocations. This is a form of optimisation because getting the current command is a common action but so far the only available option was GetUnitCommands(unitID, 1) which still allocated 3 tables.

### DIFF
--- a/rts/Lua/LuaSyncedRead.cpp
+++ b/rts/Lua/LuaSyncedRead.cpp
@@ -4271,10 +4271,10 @@ int LuaSyncedRead::GetUnitCurrentCommand(lua_State* L)
 
 	const CFactoryCAI* factoryCAI = dynamic_cast<const CFactoryCAI*>(commandAI);
 	const CCommandQueue* queue = (factoryCAI == nullptr)? &commandAI->commandQue : &factoryCAI->newUnitCommands;
-	if (queue.empty())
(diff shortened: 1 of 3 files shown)


rts/Lua/LuaSyncedRead.cpp
```diff
@@ -4258,6 +4259,32 @@ static void PackCommandQueue(lua_State* L, const CCommandQueue& commands, size_t
...
+	if (unit == nullptr)
+		return 0;
+
+	const CCommandAI* commandAI = unit->commandAI;
+	if (commandAI == nullptr)
+		return 0;
+
+	const CFactoryCAI* factoryCAI = dynamic_cast<const CFactoryCAI*>(commandAI);
+	const CCommandQueue* queue = (factoryCAI == nullptr)? &commandAI->commandQue : &factoryCAI->newUnitCommands;
+	if (queue.empty())
```
ashdnazg: will fail compilation

ashdnazg:
ty

PR #438, opened 2019-03-20
Fix #1382, draw calls see too much under ceasefire

sprunk:


ashdnazg:
I suspect you haven't tested it locally, and this is the kind of change that might have loads of unintended consequences, so until someone tests this, the PR will wait

ashdnazg:
The engine seems quite in conflict with itself whether you should or shouldn't see units of ceasefired allyteams:
https://github.com/spring/spring/blob/develop/rts/Sim/Units/Unit.cpp#L1595

sprunk:
This PR just makes LuaGL calls consistent with regular unit rendering by making it only look at the LOS_INLOS bit. Whether the engine is inconsistent about setting that bit is orthogonal IMO.

ashdnazg:
so it should just remove the entire allycheck variable...

PR #442, opened 2019-03-28
Get*Info(id, false) -> no customkeys table alloc

sprunk:


rtri:
Why aren't you pushing `nil` in `GetPlayerInfo` as well?

sprunk:
Why? That would break every existing widget that uses either the table or the multiplier.

rtri:
1) API consistency with `GetPlayerInfo`
2) anyone wanting to take advantage of this (which should be everybody maintaining an active project) will have to adjust calls regardless, two birds one stone etc

sprunk:
It is done.

rtri:
ty

PR #444, opened 2019-04-02
fix some null checks

sprunk:
other functions above (eg `IsEditDefsEnabled`) don't do a null check at all, probably either those should get one or the ones I touched here removed


rtri:
Neither `gs` nor `gameSetup` can ever be null, these checks are deadweight.

PR #447, opened 2019-04-08
Non-manualfire weapons: only 1 salvo with META

sprunk:
META doesn't seem to do anything for regular weapons but comments seem to imply it should work as a "shoot just 1 salvo".

### DIFF
--- a/rts/Sim/Units/CommandAI/CommandAI.cpp
+++ b/rts/Sim/Units/CommandAI/CommandAI.cpp
@@ -1648,11 +1648,12 @@ void CCommandAI::WeaponFired(CWeapon* weapon, const bool searchForNewTarget)
 
 	if (searchForNewTarget) {
 		// manual fire or attack commands with meta will only fire a single salvo
-		// noAutoTarget weapons finish an attack commands after a
-		// salvo if they have more orders queued
-		if (weapon->weaponDef->manualfire && !(c.GetOpts() & META_KEY))
+		if (weapon->weaponDef->manualfire && !(c.GetOpts() & META_KEY)
+		|| !weapon->weaponDef->manualfire &&  (c.GetOpts() & META_KEY))
 			orderFinished = true;
 
+		// noAutoTarget weapons finish an attack commands after a
+		// salvo if they have more orders queued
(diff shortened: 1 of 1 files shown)


rtri:
META handles "target locking", which is a property of attack and manualfire commands.

If a manualfire weapon fired and there is *no* lock, the order can be cleared.
If a regular weapon fired and there *is* a lock (your change), does finishing the order make sense?

sprunk:
Right, I didn't realize there's a builtin engine target lock (ZK reimplements something similar on its own through `Spring.SetUnitTarget`). The comments near the spot I touched imply that `META` is for 1-salvo so should probably be changed regardless of the below.

My aim here is to have a command that makes the unit shoot one salvo and finishes. This sounds a lot like `CMD_MANUALFIRE` so my initial idea was to just add `canManualFire` to the unitdef, but it turns out that this still requires the unit to have a `manualfire` weapon (I tested with ZK `vehheavyarty` and it just drove over to the target spot). I still want the weapon to be able to target stuff on its own so I don't want to add the `manualfire` tag to the weapon. I don't think I want `CMD_MANUALFIRE` affecting non-`manualfire` weapons either (it wouldn't mess much with ZK units but might with other games' and is inelegant), at least by default (a `shootAllWeaponsEvenRegularOnesOnManualFireCommand` unitdef tag or a `notReallyAManualFireWeaponButShootItOnManualFireCommandAnyway` weapondef tag would solve the problem without screwing up other games' unit behaviour). I don't know whether these would be good design so I figured I'd just "fix" (based on the obsolete comments) the META key for regular attack.

rtri:
This code was pulled in different directions between 23fae46c3db9a9c53b8446be843e1c564aa48817 and f214bad181d79db6fc57aec40c518f29bbbef119, the groundtruth is no longer obvious.

For your use-case a `WeaponFired` callin (triggered if the META bit is set) seems like the easiest road to take.

sprunk:
Alright, thanks for the tip. This sounds like something I can already do with existing interfaces (synced Lua already knows about a unit ending a salvo through LUS `EndBurst`).

PR #449, opened 2019-04-10
CmdDesc cache overflow: dump the whole cache once

sprunk:


rtri:
I have something else lined up for this.

PR #451, opened 2019-04-15
DemoTool: more verbose NETMSG_MAPDRAW

mvollrath:
The text players enter into map markers may be of interest.

Print all the other stuff too, because why not.


rtri:
Looks useful enough, thanks.

PR #452, opened 2019-04-24
Expose `Script.IsEngineMinVersion` to `defs.lua` and `modrules.lua` parsing

sprunk:


sprunk:
Ideally this specific callout would be available literally everywhere.

PR #454, opened 2019-05-01
Add `GetCurrentCamera`, tableless `GetCameraState`

sprunk:
`Spring.GetCurrentCamera` returns the name of the current camera,
for example "ta" or "spring" (currently available under the `name`
key in the table returned by `Spring.GetCameraState`).

`Spring.GetCameraState` now accepts a boolean parameter; if it is
false then instead of allocating a table, it returns cam-specific
values onto the stack (use `GetCurrentCamera` first to know how to
interpret them; use `GetCamera{Position,Direction,FOV}` to get the
non-specific values from the base class). The returned values are:
  fps, rot -> oldHeight
  ta -> height, angle, flipped
  spring -> rx, ry, rz, dist
  free -> rx, ry, rz, vx, vy, vz, avx, avy, avz
  ov -> (nil)

### DIFF
--- a/doc/changelog.txt
+++ b/doc/changelog.txt
@@ -144,6 +144,13 @@ Lua:
  ! Spring.GetTeamInfo: switched the last two return values around (incomeMultiplier is now 7th, customKeys is now 8th)
  - Added a second boolean parameter to Spring.Get{Player,Team}Info, if it is false the function will not return
    the (now) last return value, the customKeys table (allocation optimisation)
+ - Added Spring.GetCurrentCamera, returns the name of the current camera
+ - Added a boolean parameter to Spring.GetCameraState, if set to false it will return the following camera-specific values on the stack:
+     fps, rot -> oldHeight
+     ta -> height, angle, flipped
+     spring -> rx, ry, rz, dist
+     free -> rx, ry, rz, vx, vy, vz, avx, avy, avz
+     ov -> (nil)
 
 AI:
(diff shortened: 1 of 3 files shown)


ashdnazg:
Hmm, I suspect it could also work to just push the name before the rest of the args. 
Lua has no problem if it needs to unpack too little or too many arguments.
No strong opinion though, just for you to consider, if you want this merged as it is, it's also ok.

sprunk:
My aim with having a separate callout was to get this
```lua
local camtype = Spring.GetCurrentCamera()
if camtype == "spring" then
   local rx, ry, rz = Spring.GetCameraState(false)
   -- some logic
elseif camtype = "ta" then
   local height, angle, flipped = Spring.GetCameraState(false)
   -- some logic
end
```
rather than this

```lua
local camtype, p1, p2, p3 = Spring.GetCameraState(false)
-- crappy var names, no idea what they mean yet
if camtype == "spring" then
  -- some logic where p1-3 mean rotation x/y/z
elseif camtype == "ta" then
  -- some other logic in which p1-3 don't mean rotation x/y/z but height/angle/flipped
end
```

ashdnazg:
I know, I just wanted to verify you thought of the second option, since afaik your aim here is optimisation, and 1 call is less than 2 calls.

sprunk:
I added the camera name to `GetCameraState(false)`. I kept `GetCurrentCamera` but am ambivalent about it now.

ashdnazg:
haha, never knew design by committee can happen with 2 people

PR #455, opened 2019-05-07
Waypoint skipping fix for Mantis 6206.

GoogleFrog:
This PR changes the logic in CanSetNextWayPoint and splits allowSkip into allowSkip and facePoint.

Previously, CanSetNextWayPoint could return true if (allowSkip || rangeTest || facePoint). This caused waypoints to be skipped even if there was no free path to it's goal (!rangeTest) if it happened to be facing its goal (facePoint).

This change has CanSetNextWayPoint able to return true if (allowSkip || (rangeTest && facePoint)), which fixes the issue with the previous behaviour.


GoogleFrog:
An example: https://youtu.be/r_yL0HT1PPA

rtri:
Thanks. The logic seems sounder, but expect to do at least one followup.

PR #456, opened 2019-05-08
`GetUnit{Armored,IsActive}`: allow enemies in los

sprunk:
Most existing games give visual feedback regarding both states so I think this is a preferable default behaviour.


ashdnazg:
Can you changelog that please?
It's a small change, but still a change from previous behaviour

ashdnazg:
I think the mass part is missing in the code?

sprunk:
#371

ashdnazg:
Oh nice!

PR #458, opened 2019-05-24
Added options for PCF terrain shadows

lhog:
Should it be ported to maintenance too?


lhog:
I guess PCF should enjoy the same trick from PCSS to scale shadowmap sampling according to zoom.
Contrary to what's usually seen on units, terrain shadows change softness on zoom is very apparent.

lhog:
Closing due to no interest

PR #460, opened 2019-06-08
Solid objects interpolation based on prevPos instead of speed

GoogleFrog:
This PR replaces the draw interpolation based on speed with interpolation based on the units position at the start of the frame. The motivation is to better handle cases where a units speed is a poor predictor of where it will be next frame (such as when colliding with other units or terrain). This change to interpolation reduces the amount of jittering seen in these cases.

Added arguments to Set{Unit,Feature}Position and Set{Unit,Feature}Physics that reset the interpolation position. This allows lua to implement mechanics that cause discontinuous jumps in unit position, such as teleportation, without breaking the interpolation.

This PR only affects solid objects since speed is a good predictor for the next position of cegs and projectiles.

See: https://youtu.be/nQLdJUsbsRY

### DIFF
--- a/rts/Sim/Features/Feature.cpp
+++ b/rts/Sim/Features/Feature.cpp
@@ -464,10 +464,10 @@ void CFeature::ForcedMove(const float3& newPos)
 	quadField.RemoveFeature(this);
 
 	const float3 oldPos = pos;
-	prevPos = pos;
 
 	UnBlock();
 	Move(newPos - pos, true);
+	DisjointInterpolation();
 	Block();
 
 	// ForcedMove calls might cause the pstate to go stale
@@ -550,6 +550,7 @@ bool CFeature::UpdateVelocity(
 bool CFeature::UpdatePosition()
 {
 	const float3 oldPos = pos;
+	prevPos = pos;
 	// const float4 oldSpd = speed;
 
 	if (moveCtrl.enabled) {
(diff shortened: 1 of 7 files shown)


rts/Sim/Features/Feature.cpp
```diff
@@ -463,6 +464,7 @@ void CFeature::ForcedMove(const float3& newPos)
 	quadField.RemoveFeature(this);
 
 	const float3 oldPos = pos;
+	prevPos = pos;
```
ashdnazg: why don't you call DisjointInterp here?

rts/Sim/Units/UnitHandler.cpp
```diff
@@ -321,6 +321,11 @@ void CUnitHandler::UpdateUnitMoveTypes()
 {
 	SCOPED_TIMER("Sim::Unit::MoveType");
 
+	for (activeUpdateUnit = 0; activeUpdateUnit < activeUnits.size(); ++activeUpdateUnit) {
```
ashdnazg: I don't think it's good to do this in a separate loop, you'll get a zillion cache misses.

rts/Sim/Features/Feature.cpp
GoogleFrog: That is a good point. I also did not update prevPos in the correct movement function.

rts/Sim/Units/UnitHandler.cpp
GoogleFrog: That is an issue. Units can collide with each other throughout the main loop so I'm not keen to set prevPos for each one individually. I can try it out though.

GoogleFrog:
I've removed the extra loop. The interpolation still looks ok.

rtri:
I'm not sure paying 12 extra bytes per object (including features, which rarely ever move) for interpolation is the best deal here.

GoogleFrog:
I am not sure either. The feature interpolation can be removed though.

PR #462, opened 2019-06-16
Add `Spring.UnitFinishCommand(unitID)`

sprunk:
Finishes current command.
 * similar to `CMD.REMOVE`, but puts the command back onto the queue with the Repeat state.
 * similar to returning `true, true` from `CommandFallback` but works on builtin engine commands.


sprunk:
See #463 for an alternate approach.

PR #463, opened 2019-06-18
`CMD.REMOVE`: `META` option now means finish

sprunk:
Behaves as if the command finished normally (ie. gets put back at the end of the queue if Repeat state is enabled)

An alternate approach to #462.


rtri:
CAI code is already too convoluted and META has quite a few overloads for user commands, so let's go with the explicit callout.

PR #471, opened 2019-08-17
Better fix(?) to https://springrts.com/mantis/view.php?id=6100

lhog:
Hopefully fixed bump water flickering by making znear larger and introducing glPolygonOffset into the bump water rendering loop. znear increase doesn't seem to reproduce any ill effects during my tests.

maintenance version for review and tests.


rtri:
Thanks, it should be possible to produce tighter n/f bounds when looking at the ground but this will work fine too.

PR #472, opened 2019-08-18
Added map compiler as tool

Oflameo:
This pull request is to fix the issue https://github.com/spring/SpringMapConvNG/issues/2 and to make it easier to start making maps.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 [submodule "tools/pr-downloader"]
 	path = tools/pr-downloader
 	url = git://github.com/spring/pr-downloader.git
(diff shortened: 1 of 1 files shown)


ashdnazg:
I think you have to add it to cmake as well and distribution/automatic compilation is another matter that has to be dealt with

Oflameo:
Okay; Take 2!

abma:
hint: https://github.com/Oflameo/spring/blob/b22445ce83d58472aeee11a19adfa2226ab4fce0/tools/CMakeLists.txt#L14

abma:
FYI: https://springrts.com/phpbb/viewtopic.php?t=40175

PR #475, opened 2019-09-30
update circuitai

abma:
+ fix compile warning


ashdnazg:
Sorry, only saw that after pushing

PR #476, opened 2019-10-02
Seems like a fix to https://springrts.com/mantis/view.php?id=6307

lhog:
White cutoff wall as per https://springrts.com/mantis/view.php?id=6307 seems to have been caused by:
*  `std::max(0.0f, ...)` of `std::max(0.0f, forward.dot(UpVector)))` when camera view angle becomes almost parallel to the ground
*  and insufficient default maxViewRange in the rest of the cases

### DIFF
--- a/rts/Game/Camera.cpp
+++ b/rts/Game/Camera.cpp
@@ -278,7 +278,7 @@ void CCamera::UpdateViewRange()
 	// camera-height dependence (i.e. TAB-view)
 	wantedViewRange = std::max(wantedViewRange, (pos.y - std::max(0.0f, mapMinHeight)) * 2.0f);
 	// view-angle dependence (i.e. FPS-view)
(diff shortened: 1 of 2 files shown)


rtri:
>  when camera view angle becomes almost parallel to the ground

In this case `forward.dot(UpVector)` will approach 0, how does removing the `std::max` help?

Also prefer to keep maxViewRange small(er): `MAX_VIEW_RANGE * 0.6f` is 39K elmos which seems excessive.

lhog:
No idea, but exactly this line `std::max(0.0f, ...)` was the difference between older code and the new one. Also I changed exactly this one line and white cutoff, caused by the steep camera angle, has gone away.

As far as maxViewRange, I logged the wantedViewRange in the different zoom and angle settings and the maximum value I observed was around 38000 elmo, thus `0.6f * 65535.0f`

PR #481, opened 2019-11-03
Prevent nullpo in LuaMenu Reload call

esainane:
`game` can be a null pointer when LuaMenu calls `SpringApp::Reload`.

In practice this didn't cause many problems, as this gets optimised as
a fixed function call, with a null `this` parameter. This led it to be
left unnoticed.

However, when using tools that are stricter about bad behaviour
(-fsanitize=...), this would detect a null pointer dereference and
abort the program.

Encountered incidentally during mantis #6327.


rtri:
tyvm

PR #484, opened 2019-11-09
Fail-fast on missing headers

hoijui:
Fail in CMake config phase, instead of later, when compiling.

Tested on *Debian testing*.

... hey folks! :-) ;-)


abma:
fontconfig support is optional:

what/where did it fail to compile?

gajop:
It seems that this PR detects when misconfigured/misinstalled versions of fontconfig and freetype are found on the system, by validating if they have the proper headers.
Is this something we should really be doing though? I'm having trouble understanding a legitimate situation in which such a thing can happen.

Also hey @hoijui , hope you're doing well :)

hoijui:
if this is what it does, then I would recommend to issue a warning that sais exactly that.
why would we not make life easier for a user, even if it is an edge-case?
it can safe a lot of bug hunting time, even if it just happens once in springs history (and it already did happen at least once ;-) )

.. hey gajop and abma! :-)

btw... to you and the other folks, next time you talk to them: I am now working with opensourceecology.de (we try to push open hardware documentation and standards, and coordinate hardware project makers and the like).
.. in case any of you are interested in that stuff, ping me! :-)

.. I would say we (German section) are not as cool as the US section, which are also much more famous, but we have some hot stuff too. :-)

abma:
thank you!

PR #487, opened 2019-12-16
[AI] Fix Command_getParams

rlcevg:
LegacyCpp AI interface wrapper is written in a way that expects real array size as return value after [1st call](https://github.com/spring/spring/blob/39f8fbd47f112f067999366d0444fa591480f9b3/AI/Wrappers/LegacyCpp/AIAICallback.cpp#L295) with outArray=nullptr. The issue could be fixed on LegacyCpp level (pass INT_MAX instead of 0), but then in general C interface becomes inconsistent with similar functions like [getLosMap](https://github.com/spring/spring/blob/39f8fbd47f112f067999366d0444fa591480f9b3/rts/ExternalAI/SSkirmishAICallbackImpl.cpp#L1946-L1957).
(Fixes KAIK, please populate to maintenance)

### DIFF
--- a/rts/ExternalAI/SSkirmishAICallbackImpl.cpp
+++ b/rts/ExternalAI/SSkirmishAICallbackImpl.cpp
@@ -3512,16 +3512,19 @@ EXPORT(int) skirmishAiCallback_Unit_SupportedCommand_getParams(
 	/*const*/ CAICallback* cb = GetCallBack(skirmishAIId);
 	const std::vector<const SCommandDescription*>* cq = cb->GetUnitCommands(unitId);
 	const std::vector<std::string>& ps = cq->at(supportedCommandId)->params;
-
 	const int paramsRealSize = ps.size();
 
+	int paramsSize = paramsRealSize;
+
 	if (params != nullptr) {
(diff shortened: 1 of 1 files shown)


rlcevg:
There is strange difference between develop and maintenance
~~~
$ git diff develop..maintenance -- rts/ExternalAI/SSkirmishAICallbackImpl.cpp
diff --git a/rts/ExternalAI/SSkirmishAICallbackImpl.cpp b/rts/ExternalAI/SSkirmishAICallbackImpl.cpp
index ee1eff72e7..41d50cd166 100644
--- a/rts/ExternalAI/SSkirmishAICallbackImpl.cpp
+++ b/rts/ExternalAI/SSkirmishAICallbackImpl.cpp
@@ -836,7 +836,9 @@ EXPORT(int) skirmishAiCallback_Engine_handleCommand(
                } break;
 
                case COMMAND_DRAWER_PATH_FINISH: {
-                       // noop
+                       //const SFinishPathDrawerCommand* cmd =
+                       //              (SFinishPathDrawerCommand*) commandData;
+                       clb->LineDrawerFinishPath();
                } break;
~~~
Probably gfx things and hence "noop" in develop

rlcevg:
do not merge, updating

rlcevg:
done editing

PR #489, opened 2019-12-18
Rifle weapontype: remove hardcoded particles

sprunk:
I don't think anybody is using Rifles (harcoded low-quality particles being one of the reasons!) so this should do no harm and is fairly easy to replicate with Lua. On the other hand, Rifle could become more useful as a very simple hitscan weapon (simpler than BeamLaser which still involves a projectile).

### DIFF
--- a/rts/Sim/Weapons/Rifle.cpp
+++ b/rts/Sim/Weapons/Rifle.cpp
@@ -5,13 +5,9 @@
 #include "Game/TraceRay.h"
 #include "Game/GameHelper.h"
 #include "Map/Ground.h"
-#include "Rendering/Env/Particles/Classes/HeatCloudProjectile.h"
-#include "Rendering/Env/Particles/Classes/SmokeProjectile.h"
-#include "Rendering/Env/Particles/Classes/TracerProjectile.h"
 #include "Sim/Misc/GlobalSynced.h"
 #include "Sim/Units/Unit.h"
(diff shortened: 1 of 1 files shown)


silentwings:
iirc rifles are used for snipers in some *A games

sprunk:
Not in BA: https://github.com/Balanced-Annihilation/Balanced-Annihilation/blob/master/units/ArmKbots/T2/armsnipe.lua#L154

Not in BAR: https://github.com/beyond-all-reason/Beyond-All-Reason/blob/master/units/ArmKbots/T2/armsnipe.lua#L157

Not in TechA: https://github.com/nixtux/TA/blob/master/units/armsnipe.lua#L122

Not in ZK: https://github.com/zerok-rts/zero-k/blob/master/units/cloaksnipe.lua#L114

sprunk:
Not BA(R).

I can't find a TechA repo, I found https://github.com/n3wm1nd/TA which does have Rifles but that one was last updated in 2016 so would already be broken by the myriad engine changes since then.

abma:
techa is at https://github.com/techannihilation/TA

sprunk:
Alright, so TechA uses Rifle - it would probably be good to wait with this patch for goodwill reasons then. At some point I might submit a patch to make Rifle support CEGs which would make it easier to replicate those particles.

PR #492, opened 2020-02-11
Added stopToAttack 

GoogleFrog:
The PR adds a tag to disable MobileCAI giving move orders while attacking. It is a step towards making polishing starcraft-style unit interactions.

Without this change units set to maneuver or roam move along with their target, even when in range.I do not want to use hold position because it removes idle target acquisition and affects fight.

This is a compatible changed cherry picked from the fork I referenced in this ticket: https://springrts.com/mantis/view.php?id=6199

### DIFF
--- a/rts/Lua/LuaUnitDefs.cpp
+++ b/rts/Lua/LuaUnitDefs.cpp
@@ -865,6 +865,7 @@ ADD_BOOL("canAttackWater",  canAttackWater); // CUSTOM
 
 	ADD_BOOL("levelGround", ud.levelGround);
 	ADD_BOOL("strafeToAttack", ud.strafeToAttack);
+	ADD_BOOL("stopToAttack", ud.stopToAttack);
 
 	ADD_BOOL( "useBuildingGroundDecal",  ud.decalDef.useGroundDecal);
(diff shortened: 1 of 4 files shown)


ashdnazg:
Since units take time to aim you will have units start-stopping on the edge of the range not being able to fire at units slower than them.
Did you test this?

GoogleFrog:
I did not test that specifically, but it is what I would expect happens for units with high deceleration and aim time. Here is a video of a unit with stopToAttack, along with other things in its script that enforce only firing while stopped: https://www.youtube.com/watch?v=w5hD4-eJzp0

If aim time is low enough then the unit should be able to fire. I expect that a game could remedy the situation by making their units get closer to fire against fleeing units. Other games with stopToAttack seem to get around this by making units able to shoot at things even if they have just left range.

gajop:
> I expect that a game could remedy the situation by making their units get closer to fire against fleeing units. Other games with stopToAttack seem to get around this by making units able to shoot at things even if they have just left range

Any plans to add either of these things to the engine with this PR? Otherwise I think this behavior works well and is simple enough to understand so I approve.

GoogleFrog:
Not with this PR.

gajop:
OK, add it to the changelog and I'll merge

PR #493, opened 2020-02-11
Added pathRadiusMult.

GoogleFrog:
This PR adds and exposes a multiplier to the sphere of repulsive force between units. This is part of my attempt to fix pushResistant and give Spring the parameters required to make more starcraft-style units. A multiplier of around 1.3 does not appear to break pathfinding and makes units less clumped. Simply increasing their footprint is a problem with pushResistant, since it blocks and area based on footprint size.

If adding an extra float to SolidObject is somehow terrible for performance then I'd consider adding it to movedefs instead.

### DIFF
--- a/rts/Sim/MoveTypes/GroundMoveType.cpp
+++ b/rts/Sim/MoveTypes/GroundMoveType.cpp
@@ -402,7 +402,7 @@ CGroundMoveType::CGroundMoveType(CUnit* owner):
 	// unit-gravity must always be negative
 	myGravity = mix(-math::fabs(ud->myGravity), mapInfo->map.gravity, ud->myGravity == 0.0f);
 
-	ownerRadius = md->CalcFootPrintMinExteriorRadius();
+	ownerRadius = owner->CalcFootPrintMinExteriorRadius();
 }
 
 CGroundMoveType::~CGroundMoveType()
@@ -1298,7 +1298,7 @@ float3 CGroundMoveType::GetObstacleAvoidanceDir(const float3& desiredDir) {
 	// avoider always uses its never-rotated MoveDef footprint
 	// note: should increase radius for smaller turnAccel values
 	const float avoidanceRadius = std::max(currentSpeed, 1.0f) * (avoider->radius * 2.0f);
-	const float avoiderRadius = avoiderMD->CalcFootPrintMinExteriorRadius();
+	const float avoiderRadius = avoider->CalcFootPrintMinExteriorRadius();
 
 	QuadFieldQuery qfQuery;
 	quadField.GetSolidsExact(qfQuery, avoider->pos, avoidanceRadius, 0xFFFFFFFF, CSolidObject::CSTATE_BIT_SOLIDOBJECTS);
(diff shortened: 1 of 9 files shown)


rts/Sim/MoveTypes/GroundMoveType.cpp
```diff
@@ -1761,8 +1761,8 @@ void CGroundMoveType::HandleObjectCollisions()
 		// NOTE:
 		//   use the collider's MoveDef footprint as radius since it is
 		//   always mobile (its UnitDef footprint size may be different)
-		const float colliderFootPrintRadius = colliderMD->CalcFootPrintMaxInteriorRadius(); // ~= CalcFootPrintMinExteriorRadius(0.75f)
-		const float colliderAxisStretchFact = colliderMD->CalcFootPrintAxisStretchFactor();
+		const float colliderFootPrintRadius = collider->CalcFootPrintMaxInteriorRadius(); // ~= CalcFootPrintMinExteriorRadius(0.75f)
```
sanguinariojoe: I would say this is not a good business at all... Look the comment...

The path finder never will use the unit footprint. If the unit footprint is smaller than the corresponding movedef one, then it would be perfectly feasible to get stucked units. i.e. They are parking close together, and then pathfinding is classifying them as blocked, not letting them to move anymore

rts/Sim/Objects/SolidObject.cpp
```diff
@@ -448,8 +449,8 @@ void CSolidObject::Kill(CUnit* killer, const float3& impulse, bool crushed)
 
 
 
-float CSolidObject::CalcFootPrintMinExteriorRadius(float scale) const { return ((math::sqrt((xsize * xsize + zsize * zsize)) * 0.5f * SQUARE_SIZE) * scale); }
-float CSolidObject::CalcFootPrintMaxInteriorRadius(float scale) const { return ((std::max(xsize, zsize) * 0.5f * SQUARE_SIZE) * scale); }
+float CSolidObject::CalcFootPrintMinExteriorRadius(float scale) const { return ((pathRadiusMult * math::sqrt((xsize * xsize + zsize * zsize)) * 0.5f * SQUARE_SIZE) * scale); }
```
sanguinariojoe: Since you are just multiplying the `math::sqrt((xsize * xsize + zsize * zsize))` by a factor, why is this different from just simply increasing the footprint?

GoogleFrog:
Have you tested it to check whether small multipliers break pathfinding? Units pathing around each other with push resistant has been broken for a long time and this is an incremental change towards making it work.

sanguinariojoe:
I'm not even discussing about the value of the multiplier. The problem I see (probably I'm missing something) is that you are changing the footprint used to compute the collision from the movedef one to the unitdef one. But AFAIK the pathfinding is just considering the movedef one (because using the unitdef one is unfeasible). 

So 2 things may happens:

* Unitdef and movedef footprints are significantly different. If the unitdef footprint is smaller, the units can stick so close than afterwards pathfinding will detect both as stuck, right?
* Unitdef footprint is never too different from movedef footprint. Then the multiplier has basically the same effect than increasing the footprint, right?

GoogleFrog:
Setting unitdef and movedef footprint to different values is a bad idea modside for other reasons. But you're right, I'll revisit this later with movedef size on that line.

gajop:
> Setting unitdef and movedef footprint to different values is a bad idea modside for other reasons. But you're right, I'll revisit this later with movedef size on that line.

I assume this PR is still waiting a revision and I don't have to look at it yet?

GoogleFrog:
Yes. I don't know why unitdef and movedef footprints would ever not match, but the switch to movedef footprint (if the unit has a movedef) may as well be done.

GoogleFrog:
Whatever. I think he was being overly cautious rather than testing his concerns, but I'll be PRing to BAR if I reopen this.

PR #494, opened 2020-02-18
Add a single threaded ROAM without dummy overhead

GoogleFrog:
I have been investigating the performance issues raised in https://springrts.com/mantis/view.php?id=6362 and https://springrts.com/mantis/view.php?id=6340. 

Through building various versions and benchmarking them I have concluded that the threading in ROAM has little benefit. This benchmark shows that ROAM became slower after https://github.com/spring/spring/commit/4bc8b64704f4036638b10a91e2dd46cb836323b0 due to the additional overhead of the dummies, not because threading was removed. The benchmark involves moving the camera rapidly around a unitless map.
image

This PR adds an additional meshdrawer based on the parent of https://github.com/spring/spring/commit/4bc8b64704f4036638b10a91e2dd46cb836323b0. The access violation identified by Krogoth is resolved by removing the concurrent patch updates. Doing so appears to have no impact on performance. The change here is compatible with future work on the threaded ROAM, since the threaded ROAM is unaffected, and furthermore may be used as a comparison. The unthreaded ROAM should allow games to use the latest engine versions until kloot, or anyone else, has more time to work on terrain rendering.

Here is a benchmark of the PR compared to a few other versions. It is not ideal because it mixes buildbot and locally build engines. This is what the benchmark looks like: https://www.youtube.com/watch?v=2tpk9b6ITFI
image
I previously thought the performance issue with 1463 had been fixed, however that benchmark did not have any camera movement. The small amount of camera movement in the updated benchmark reveals the issue with 1463.

Here is a video comparing the meshdrawers: https://www.youtube.com/watch?v=wZH-g1RiDTY


gajop:
I haven't looked into reviewing the code but I have a comment on the design.

This change introduces (a slightly?) different behavior by copying original classes and modifying some of its code. The new behavior is then available as a switch and would be used some games, probably ZK.

The problem with this is that it makes code maintenance much more difficult as any future changes would likely have to happen in 2 places (4 if we include develop and maintenance branches) and that's  just hard to track. The other issue is that we'd likely split our userbase and one of the methods will just not be tested much.

My preference would really be to just overwrite or extend the existing implementation. Have a single *best* method, especially since there's probably no big difference in the techniques

GoogleFrog:
I don't mind having this PR overwrite the existing implementation. My intention was to step on the fewest toes. As an additional benefit it makes comparing the two systems easier.

gajop:
Can you please make such a PR then please? It'll make reviewing easier too.

GoogleFrog:
If you want. Here is one: https://github.com/spring/spring/pull/497

rtri:
Just as a PSA, my position is as follows:

1) this PR exists ostensibly so ZK can move beyond 1435-g79d77ca
2) I personally neither can nor want to continue the maintenance branch anymore, other devs seem to have similar feelings and/or life constraints judging from the recent general level of [in]activity
3) as such, whether ZK stays on 1435-g79d77ca forever or on 1464-g4f428c8+PR494+? forever with the last handful of additional fixes brought by that build makes little difference afaict
4) for ROAM the proper (but difficult) way forward is to implement fully asynchronous tesselation such that its CPU footprint gets addressed even on maps like Onyx Cauldron where the algorithm suffers, dummies or no dummies
5) any future work on terrain rendering (if done by me) would instead be exclusive to develop, and certainly involve stripping out all ROAM code in favor of a more modern GPU approach
6) at this point I have no real interest in touching develop again either until the major games demonstrate a credible commitment to switch to it, which does not look to (ever) be happening as a result of excess legacy baggage
7) the above facts mean that Spring (the engine at least) is at a standstill currently unless someone else not only radically disagrees but has the time/energy/skill/etc to both restore inertia and deal with the develop/maintenance split
8) under these circumstances, I don't think it makes sense for me to merge any pull request other than perhaps a fix for some critical exploit

The intention here is appreciated, but we're probably past the point of no return.

abma:
@rtri 

does this mean, current development should be usable for games with major changes in lua code?

if so, can you give some hints where/how to start?

GoogleFrog:
@rtri that makes sense. The only relevant disagreement I have is that I also made these PRs with BAR in mind, since they are interested in using develop and it is sad to see them using the legacy mesh drawer.

lhog:
@rtri I'm actually working on adding GPU tesselation based mesh drawer. For now I target it for `maintenance` (it will only work for drivers, with the required extensions), but it should easily be portable to `develop` as well.

@abma too bad you seem to be not on Discord, there is a lot of interesting talks going on regarding `develop`. With @rtri being currently inactive, I think I'm the 2nd best person to talk to w.r.t. moving to the games to `develop`. The problem is, despite the obvious increase in performance and possibility to put together the new graphical effects/systems efficiently, the amount of work to port a game itself is quite big and the game might even lose some players due to much higher OpenGL version requirements. Anyway, to answer your question, folks who would like to start the journey should feel free to approach me.

abma:
@lhog:

> Anyway, to answer your question, folks who would like to start the journey should feel free to approach me.

so, is develop "ready" to be used? does it make sense to create a new stable release of the develop branch?

lhog:
@abma it wss not completely finished last time I checked, but I would say it is 90-95% complete and certainly is complete enough to start porting the games.

gajop:
In light of the recent merge, are any of these changes still wanted?

ashdnazg:
no

GoogleFrog:
Read the ticket: https://springrts.com/mantis/view.php?id=6340

Beherith's patch from 5 days ago doesn't seem to fix the issue.

PR #496, opened 2020-02-26
fix SetAdvantage function to allow negative values

Yaribz:
it seems no one tried to use player handicap in nearly 10 years

(discussion [here](https://github.com/spring/spring/commit/95b295b0ca07c6c08463a04a0d1d800ebbdf3e21#r37501043))


gajop:
seems good, thanks!

PR #497, opened 2020-03-12
Removed MT from ROAM

GoogleFrog:
This is a ROAM revert to the version prior to https://github.com/spring/spring/commit/4bc8b64704f4036638b10a91e2dd46cb836323b0. See https://github.com/spring/spring/pull/494

This has been compiled and tested. It seems to have the same performance improvement as the previous PR.


GoogleFrog:
Conflicts, see https://github.com/spring/spring/pull/494 and https://github.com/spring/spring/pull/495.

PR #498, opened 2020-03-21
Added the modrule decloakRequiresLineOfSight

GoogleFrog:
This controls whether sight distance is used as a filter on proximity cloak checks, allowing for cheap implementations of the behaviour desired by FLOZi for s44 and myself for ZK.
Added a missing comment for alwaysVisibleOverridesCloaked.

See:
 * https://springrts.com/mantis/view.php?id=6306 - Behaviour change requested by FLOZi.
 * https://github.com/spring/spring/commit/ca1557cdee3fe4075ff9d987dfe0926a229146d6 - Behaviour change implemented.
 * https://springrts.com/mantis/view.php?id=6374 - My ticket when I noticed that the behaviour changed.
 * http://zero-k.info/Forum/Thread/29941 - Thread reporting the change.

### DIFF
--- a/rts/Sim/Misc/ModInfo.cpp
+++ b/rts/Sim/Misc/ModInfo.cpp
@@ -85,6 +85,7 @@ void CModInfo::ResetState()
 
 		requireSonarUnderWater = true;
 		alwaysVisibleOverridesCloaked = false;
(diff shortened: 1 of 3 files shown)


gajop:
Seems good to me. thanks!

PR #500, opened 2020-03-23
Fix crash when modInfo.allowSepAxisCollisionTest = true

sanguinariojoe:
`collideeMD` can be a `nullptr`, as it is stated in https://github.com/spring/spring/blob/develop/rts/Sim/MoveTypes/GroundMoveType.cpp#L2008:

```
		const bool collideeMobile = (collideeMD != nullptr); // maybe true
```

And indeed it happens with every single unit which does not needs SAT, right after building it in a factory. Thus, calling to `collideeMD->CalcFootPrintAxisStretchFactor()` is making the engine crash.


sanguinariojoe:
Wrong branch...

PR #501, opened 2020-03-23
Fixed invalid collideeMD usage when modInfo.allowSepAxisCollisionTest = true (maintenance)

sanguinariojoe:
collideeMD can be a nullptr, as it is stated in https://github.com/spring/spring/blob/develop/rts/Sim/MoveTypes/GroundMoveType.cpp#L2008:

		const bool collideeMobile = (collideeMD != nullptr); // maybe true

And indeed it happens with every single unit which does not needs SAT, right after building it in a factory. Thus, calling to collideeMD->CalcFootPrintAxisStretchFactor() is making the engine crash.

### DIFF
--- a/rts/Sim/MoveTypes/GroundMoveType.cpp
+++ b/rts/Sim/MoveTypes/GroundMoveType.cpp
@@ -2038,7 +2038,7 @@ void CGroundMoveType::HandleUnitCollisions(
 		const float2 collideeParams = {collidee->speed.w, collideeMobile? collideeMD->CalcFootPrintMaxInteriorRadius(): collidee->CalcFootPrintMaxInteriorRadius()};
 		const float4 separationVect = {collider->pos - collidee->pos, Square(colliderParams.y + collideeParams.y)};
 
-		if (!checkCollisionFuncs[allowSAT && (forceSAT || (collideeMD->CalcFootPrintAxisStretchFactor() > 0.1f))](separationVect, collider, collidee, colliderMD, collideeMD))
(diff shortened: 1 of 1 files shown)


rts/Sim/MoveTypes/GroundMoveType.cpp
```diff
@@ -2038,7 +2038,7 @@ void CGroundMoveType::HandleUnitCollisions(
 		const float2 collideeParams = {collidee->speed.w, collideeMobile? collideeMD->CalcFootPrintMaxInteriorRadius(): collidee->CalcFootPrintMaxInteriorRadius()};
 		const float4 separationVect = {collider->pos - collidee->pos, Square(colliderParams.y + collideeParams.y)};
 
-		if (!checkCollisionFuncs[allowSAT && (forceSAT || (collideeMD->CalcFootPrintAxisStretchFactor() > 0.1f))](separationVect, collider, collidee, colliderMD, collideeMD))
+		if (!checkCollisionFuncs[allowSAT && (forceSAT || (collideeMobile ? collideeMD->CalcFootPrintAxisStretchFactor() > 0.1f : false))](separationVect, collider, collidee, colliderMD, collideeMD)) {
```
gajop: I think it's more natural to write `collideeMobile && collideeMD->CalcFootPrintAxisStretchFactor() > 0.1f` instead of using a ternary operator with `false` constant

gajop:
This should target develop first (assuming it's not merged there yet)

sanguinariojoe:
OK, I'm making the changes in develop then!

sanguinariojoe:
* I implemented your clever suggested change
* I created the develop branch PR (https://github.com/spring/spring/pull/502)

What is the official way to ask for changes become integrated also in maintenance? (after merging in develop)

gajop:
Seems good - thanks. I'll just merge both of them directly.
I think what you did now is an acceptable approach (two PRs), but definitely having the develop one is a good first step - it can always be cherry picked manually to maintenance.

PR #503, opened 2020-03-26
Coerce to number less often in *SetRulesParam

esainane:
Unlike most lua_is* functions, lua_isnumber checks to see if the
given value *can* be a number, rather than if it currently is one:

> Returns 1 if the value at the given acceptable index is a number or
> a string convertible to a number, and 0 otherwise.

This means that even if a lua caller expressly coerces to string
beforehand, as long as the string looks like a number, it gets stored
as a number. This can cause problems if, for example, a player name
that looks like a very large number loses precision.

See esainane/zkstats@99ddb35 for an example.

Explicitly checking using the lua_israwnumber variant allows the caller
to provide either a string or a number, and have it be stored as such.

### DIFF
--- a/rts/Lua/LuaSyncedCtrl.cpp
+++ b/rts/Lua/LuaSyncedCtrl.cpp
@@ -1030,7 +1030,7 @@ void SetRulesParam(lua_State* L, const char* caller, int offset,
 	LuaRulesParams::Param& param = params[key];
 
 	// set the value of the parameter
-	if (lua_isnumber(L, valIndex)) {
+	if (lua_israwnumber(L, valIndex)) {
 		param.valueInt = lua_tofloat(L, valIndex);
 		param.valueString.resize(0);
 	} else if (lua_isstring(L, valIndex)) {


rts/Lua/LuaSyncedCtrl.cpp
```diff
@@ -1030,7 +1030,7 @@ void SetRulesParam(lua_State* L, const char* caller, int offset,
 	LuaRulesParams::Param& param = params[key];
 
 	// set the value of the parameter
-	if (lua_isnumber(L, valIndex)) {
+	if (lua_type(L, valIndex) == LUA_TNUMBER) {
```
sprunk: `lua_israwnumber`?

sprunk:
```suggestion
	if (lua_israwnumber(L, valIndex)) {
```

esainane: PR updated, thank you!

ashdnazg:
please reopen to develop.

esainane:
This is now in develop, can this be cherry-picked to maintenance?

ashdnazg:
https://github.com/spring/spring/commit/ffafcbc51a0d74cdf957798002e0d4ee979c7c1c

PR #504, opened 2020-03-26
Fix ROAM bug 6340 https://springrts.com/mantis/view.php?id=6340, any change in…

Beherith:
… visibity forces subsequence patches Computevariance to be called again.

Patch::ComputeVariance only needs to be called if patch is visible and dirty. Before this change it was called even if tesselMesh was true, as the return value of |= is equal to tesselMesh. 
Thus if tesselMesh was set to true previously by a visibility change, then all subsequence patches would get their variance recomputed (which is quite expensive)


lhog:
from BAR Discord:

Before:



After:

abma:
wow, thank you!

PR #506, opened 2020-03-29
Added initial support for MCST Elbrus 2000 (e2k) CPU architecture + x86-64 fix

r-a-sattarov:
- added initial support in spring engine
- added initial support in 7z lib
- added initial support in asio lib
- added initial support in assimp lib
- fixed -march option for x86-64 arch

Added MCST Elbrus 2000 (e2k) CPU architecture initial support.
This is VLIW/EPIC architecture, like Intel Itanium.

About this architecture:
- https://en.wikipedia.org/wiki/Elbrus_2000


rts/build/cmake/TestCXXFlags.cmake
```diff
@@ -137,10 +137,15 @@ IF    (NOT MSVC AND NOT DEFINED MARCH)
...
 	if    ((CMAKE_SIZEOF_VOID_P EQUAL 8) AND (NOT MARCH))
 		# always syncs with 32bit
-		check_cxx_accepts_flag("-march=x86_64" HAS_X86_64_FLAG_)
+		check_cxx_accepts_flag("-march=x86-64" HAS_X86_64_FLAG_)
 		if    (HAS_X86_64_FLAG_)
-			set(MARCH "x86_64")
+			set(MARCH "x86-64")
 		endif (HAS_X86_64_FLAG_)
+		# MCST lcc compiler accept -march=elbrus-v2/v3/v4/v5/v6 and -march=native
+		check_cxx_accepts_flag("-march=elbrus-v2" HAS_E2K_FLAG_)
```
ashdnazg: This assumes you always want to compile to the native machine and never cross-compile.

r-a-sattarov: Yes, by default, the e2k MCST lcc compiler turn on -march with a value corresponding to the architecture of the processor on which the compiler is currently running.
The x86 MCST lcc cross-compiler turn on -march with a value corresponding to the architecture of the e2k target-processor (that was tuned by default).
Therefore, the optimal value will be "-march=native" for both cases.

ashdnazg:
thanks!

abma:
this broke compile on mingw64:

https://buildbot.springrts.com/builders/full-windows-x64/builds/3359/steps/buildConfigNinja/logs/stdio

ashdnazg:
I'll check this later

r-a-sattarov:
> I'll check this later

This behavior is caused by the fact that I changed the value of the "-march" option from "x86_64" to "x86-64" (in spring/rts/build/cmake/TestCXXFlags.cmake), because only the value "x86-64" is correct for the GCC compiler:
- https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

For this reason, most likely, the “HAS_X86_64_FLAG_” flag is not activated when compiling in mingw64 + x64.
And this leads to the fact that the -DSTREFLOP_SOFT option is used (in spring/CMakeLists.txt), and not -DSTREFLOP_SSE (as it was before).
> this broke compile on mingw64 + x64:
> 
> https://buildbot.springrts.com/builders/full-windows-x64/builds/3359/steps/buildConfigNinja/logs/stdio
> 
> "SoftFloatWrapper.h: No such file or directory"

I noticed that in the built-in streflop library there are no Soft implementation (there is only a SSE and X87 implementations).
Could you please add the missing Soft implementation? Or fully update the built-in streflop library from your repository?
- https://github.com/abma/streflop

abma:
interesting, this change was added ~8 years ago:

https://github.com/r-a-sattarov/spring/commit/afc0da2bf21d245aef045be2198941929f1e37b5

gcc 5.5 seems to not support -march=x86-64 nor -march=x86_64:
https://gcc.gnu.org/onlinedocs/gcc-5.5.0/gcc/x86-Options.html

so, at the time of the change this never worked as intended it seems... weird.

wrt streflop: in "my" repository: thats just a mirror of the original package to allow source code references i.e. https://springrts.com/mantis/view.php?id=5058 . if you there are changes needed in streflop for spring, please create a pull request to spring/spring .

r-a-sattarov:
> interesting, this change was added ~8 years ago:
> 
> [afc0da2](https://github.com/spring/spring/commit/afc0da2bf21d245aef045be2198941929f1e37b5)
> 
> gcc 5.5 seems to not support -march=x86-64 nor -march=x86_64:
> https://gcc.gnu.org/onlinedocs/gcc-5.5.0/gcc/x86-Options.html
> 
> so, at the time of the change this never worked as intended it seems... weird.

Ok, when you compile on mingw64 + x64, what is the correct architecture value there, "-march=x86-64" or "-march=x86_64"?

abma:
oh, sorry: it seems CMakeCache.txt is the cause of the error.

i've cleaned the builddir and now it seems to compile fine on the linux64 buildslave at least.

@TurBoss can you please clean the builddir of the win64 buildslave, too?

TurBoss:
> @TurBoss can you please clean the builddir of the win64 buildslave, too?

sure

TurBoss:
done

PR #507, opened 2020-03-31
LOAM (Lazy ROAM) testing PR

Beherith:
Forgive the ugliness, heres the down low:

ROAM frees up the entire triangle pool that stores the triangles of each tesselated patch (129*129 hmap pixels) and retesselates all patches whenever any in the visibility of patches occures, or if the unsynced heightmap changes (e.g.) explosion
LOAM does nothing when patches only go out of view, or when patches go into view, and a not-too-stale tesselation is aviable
when a 'stale' patch comes into view, only it is retesselated, and retesselation can affect neighbours, which is tracked and updated accordingly
when a heightmap change occurs, only that patch is retesselated, and if the neighbours are affected, that is updated accordingly
The cost: the mildly increased memory cost of keeping a larger triangle pool updated
the advantage is lag-free scrolling on a map
If the LOAM triangle pool runs out for any reason, then all the stored patches are all reset, and the whole view retesselated (exactly like it happens with ROAM on every visibility change)

### DIFF
--- a/rts/Map/SMF/ROAM/Patch.cpp
+++ b/rts/Map/SMF/ROAM/Patch.cpp
@@ -183,10 +183,15 @@ void Patch::Reset()
 	baseLeft.parentPatch = this;
 	baseRight.parentPatch = this;
 
+
+	midPos.x = (coors.x + PATCH_SIZE / 2) * SQUARE_SIZE;
+	midPos.z = (coors.y + PATCH_SIZE / 2) * SQUARE_SIZE;
+	midPos.y = readMap->GetCurrAvgHeight();
 	//Reset camera
 	lastCameraPosition.x = -10000000;
 	lastCameraPosition.y = -10000000;
 	lastCameraPosition.z = -10000000;
+	camDistanceLastTesselation = 10000000;
 }
 
 
@@ -194,6 +199,7 @@ void Patch::UpdateHeightMap(const SRectangle& rect)
 {
 	const float* hMap = readMap->GetCornerHeightMapUnsynced();
 
+	float averageHeight = 0;
 	for (int z = rect.z1; z <= rect.z2; z++) {
 		for (int x = rect.x1; x <= rect.x2; x++) {
 			const int vindex = (z * (PATCH_SIZE + 1) + x) * 3;
@@ -201,11 +207,13 @@ void Patch::UpdateHeightMap(const SRectangle& rect)
 			const int xw = x + coors.x;
 			const int zw = z + coors.y;
 
-			// only update y-coord
-			vertices[vindex + 1] = hMap[zw * mapDims.mapxp1 + xw];
+			const float height = hMap[zw * mapDims.mapxp1 + xw];
+			vertices[vindex + 1] = height;
+			averageHeight += height;
 		}
 	}
-
+	
+	midPos.y = averageHeight/((PATCH_SIZE+1)*(PATCH_SIZE+1));
 	VBOUploadVertices();
 	isDirty = true;
 }
@@ -318,7 +326,7 @@ bool Patch::Split(TriTreeNode* tri)
 			Split(tbn);
 			if (tbn->parentPatch != this)
 				tbn->parentPatch->isChanged = true;
-			
+
 		}
 	} else {
 		// edge triangle, trivial case
@@ -526,6 +534,7 @@ bool Patch::Tessellate(const float3& camPos, int viewRadius, bool shadowPass)
 
 	// MAGIC NUMBER 1: scale factor to reduce LOD with camera distance
 	camDistLODFactor  = midPos.distance(camPos);
+	camDistanceLastTesselation = camDistLODFactor;
 	camDistLODFactor *= (300.0f / viewRadius);
 	camDistLODFactor  = std::max(1.0f, camDistLODFactor);
 	camDistLODFactor  = 1.0f / camDistLODFactor;
--- a/rts/Map/SMF/ROAM/Patch.cpp
+++ b/rts/Map/SMF/ROAM/Patch.cpp
@@ -212,7 +212,7 @@ void Patch::UpdateHeightMap(const SRectangle& rect)
 			averageHeight += height;
 		}
 	}
-	
+
 	midPos.y = averageHeight/((PATCH_SIZE+1)*(PATCH_SIZE+1));
 	VBOUploadVertices();
 	isDirty = true;
@@ -521,11 +521,6 @@ void Patch::ComputeVariance()
 //
 bool Patch::Tessellate(const float3& camPos, int viewRadius, bool shadowPass)
 {
-	// Set/Update LOD params (FIXME: wrong height?)
-	float3 midPos;
-	midPos.x = (coors.x + PATCH_SIZE / 2) * SQUARE_SIZE;
-	midPos.z = (coors.y + PATCH_SIZE / 2) * SQUARE_SIZE;
-	midPos.y = readMap->GetCurrAvgHeight();
 
 	// Tessellate is called from multiple threads during both passes
 	// caller ensures that two patches that are neighbors or share a
--- a/rts/Map/SMF/ROAM/Patch.cpp
+++ b/rts/Map/SMF/ROAM/Patch.cpp
@@ -40,7 +40,7 @@ void CTriNodePool::InitPools(bool shadowPass, size_t newPoolSize)
 	for (int j = 0, numThreads = ThreadPool::GetMaxThreads(); newPoolSize > 0; j++) {
 		try {
 
-            size_t thrPoolSize =  std::max((CUR_POOL_SIZE = newPoolSize),newPoolSize); //the first pool should be larger, as only full retess uses threaded
+			size_t thrPoolSize =  std::max((CUR_POOL_SIZE = newPoolSize),newPoolSize); //the first pool should be larger, as only full retess uses threaded
 			for (int i = 0; i < numThreads; i++) {
 				if (i > 0) thrPoolSize = std::max((CUR_POOL_SIZE = newPoolSize) / numThreads, newPoolSize / 3);
 
@@ -48,7 +48,7 @@ void CTriNodePool::InitPools(bool shadowPass, size_t newPoolSize)
 				pools[shadowPass][i].Resize(thrPoolSize + (thrPoolSize & 1));
 			}
 
-      LOG_L(L_INFO, "[TriNodePool::%s] newPoolSize=" _STPF_ " thrPoolSize=" _STPF_ " (numThreads=%d shadowPass=%d)", __func__, newPoolSize, thrPoolSize, numThreads, shadowPass);
+			LOG_L(L_INFO, "[TriNodePool::%s] newPoolSize=" _STPF_ " thrPoolSize=" _STPF_ " (numThreads=%d shadowPass=%d)", __func__, newPoolSize, thrPoolSize, numThreads, shadowPass);
 			break;
 		} catch (const std::bad_alloc& e) {
 			LOG_L(L_FATAL, "[TriNodePool::%s] exception \"%s\" (numThreads=%d shadowPass=%d)", __func__, e.what(), numThreads, shadowPass);
@@ -90,7 +90,7 @@ void CTriNodePool::Resize(size_t poolSize)
 	// live outside the pool, but KISS)
 	assert((poolSize & 1) == 0);
 	assert(poolSize > 0);
-	LOG_L(L_INFO, "[TriNodePool::%s] to %i ",__func__, poolSize);
+	LOG_L(L_INFO, "[TriNodePool::%s] to " _STPF_,__func__, poolSize);
 
 	tris.resize(poolSize);
 }
@@ -179,7 +179,7 @@ void Patch::Reset()
 	baseLeft.BaseNeighbor  = &baseRight;
 	baseRight.BaseNeighbor = &baseLeft;
 
-    //Connect the base triangles to their parent
+	//Connect the base triangles to their parent
 	baseLeft.parentPatch = this;
 	baseRight.parentPatch = this;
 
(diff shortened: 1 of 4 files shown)


ashdnazg:
Noice!

rts/Map/SMF/ROAM/Patch.cpp
```diff
@@ -39,9 +39,11 @@ void CTriNodePool::InitPools(bool shadowPass, size_t newPoolSize)
 {
 	for (int j = 0, numThreads = ThreadPool::GetMaxThreads(); newPoolSize > 0; j++) {
 		try {
-			const size_t thrPoolSize = std::max((CUR_POOL_SIZE = newPoolSize) / numThreads, newPoolSize / 3);
 
+            size_t thrPoolSize =  std::max((CUR_POOL_SIZE = newPoolSize),newPoolSize); //the first pool should be larger, as only full retess uses threaded
```
gajop: here and in many other places in the code you are using spaces to indent. switch to tabs

rts/Map/SMF/ROAM/Patch.cpp
```diff
@@ -228,8 +240,12 @@ bool Patch::Split(TriTreeNode* tri)
 		return true;
 
 	// if this triangle is not in a proper diamond, force split our base-neighbor
-	if (!tri->BaseNeighbor->IsDummy() && (tri->BaseNeighbor->BaseNeighbor != tri))
+	if (!tri->BaseNeighbor->IsDummy() && (tri->BaseNeighbor->BaseNeighbor != tri)){
 		Split(tri->BaseNeighbor);
+        if (tri->BaseNeighbor->parentPatch != this){
```
gajop: for consistency with rest of code, don't use  {} when there's a single statement

rts/Map/SMF/ROAM/RoamMeshDrawer.cpp
```diff
@@ -188,17 +195,81 @@ void CRoamMeshDrawer::Update()
...
 
 	Patch::UpdateVisibility(cam, patches, numPatchesX);
 
-#define RETESSELLATE_MODE 1
+    int numPatchesVisible = 0;
+    int numPatchesEnterVisibility = 0;
+    int numPatchesEnteredSkipped = 0;
+    int numPatchesExitVisibility = 0;
+    int numPatchesRetesselated = 0;    
+    bool patchesToTesselate[numPatchesX * numPatchesY];
```
gajop: this isn't standard C++, you can't create an array of dynamic size like this.
consider:
1) using C arrays and dynamically allocating them
2) using STL constructs like vector (vector of bool can be weird), unique_ptr or deque

rts/Map/SMF/ROAM/RoamMeshDrawer.cpp
```diff
@@ -216,10 +287,85 @@ void CRoamMeshDrawer::Update()
 				pvflags[i] = 0;
 			}
 
-		#else
-
+		#elif (RETESSELLATE_MODE == 3)
+            bool isVisibleNow = p.IsVisible(cam);
+            bool wasVisible = pvflags[i];
```
gajop: both of these bools should probably be const

rts/Map/SMF/ROAM/RoamMeshDrawer.cpp
```diff
@@ -216,10 +287,85 @@ void CRoamMeshDrawer::Update()
...
+            pvflags[i] = uint8_t(p.IsVisible(cam));
+
+            // first case: a patch left visibility;
+            // do nothing, just count the number of patches in total that have done so, as they still use the trinodepools
+            // when a sufficient number of patches have left visibility, we are going to have to force a reset.
+
+            if (!isVisibleNow && wasVisible){
+                numPatchesExitVisibility ++;
```
gajop: no space here

rts/Map/SMF/ROAM/RoamMeshDrawer.h
```diff
@@ -77,6 +77,9 @@ class CRoamMeshDrawer : public IMeshDrawer, public CEventClient
 
 	float3 lastCamPos[MESH_COUNT];
 
+	int numPatchesLeftVisibility[MESH_COUNT] = {0};
+	int tesselationsSinceLastReset[MESH_COUNT] = {0};
```
gajop: don't think 0 is necessary here

rts/Map/SMF/ROAM/RoamMeshDrawer.cpp
ashdnazg: vector is the best way I think, might be worth to reuse it so you don't reallocate it each time

rts/Map/SMF/ROAM/RoamMeshDrawer.h
ashdnazg: are you sure?

gajop: In C++ `{}` zero-initializes arrays.

https://godbolt.org/z/EFMBbW 

`{ 0 } ` says "initialize the first element to 0, and zero initialize the other elements", which just opens things to confusion and may lead people into thinking all values are initialized to what is in the braces.

Which it is not: https://godbolt.org/z/FzirNb

rts/Map/SMF/ROAM/Patch.cpp
```diff
@@ -175,6 +178,15 @@ void Patch::Reset()
 	// attach the two base-triangles together
 	baseLeft.BaseNeighbor  = &baseRight;
 	baseRight.BaseNeighbor = &baseLeft;
+
+    //Connect the base triangles to their parent
+	baseLeft.parentPatch = this;
+	baseRight.parentPatch = this;
+
+	//Reset camera
+	lastCameraPosition.x = -10000000;
```
sprunk: the constant is somewhat magic

rts/Map/SMF/ROAM/RoamMeshDrawer.cpp
```diff
@@ -33,6 +33,11 @@ LOG_REGISTER_SECTION_GLOBAL(LOG_SECTION_ROAM)
 #define LOG_SECTION_CURRENT LOG_SECTION_ROAM
 
 
+//tessmode can be 1,2,(3 is LOAM)
+#define RETESSELLATE_MODE 3
+#define RETESSELATE_TO_FREE_INVISIBLE_PATCHES 2
+#define MAGIC_RETESSELATE_RETURNING_PATCH_CAMDIST 1000.0f * 1000.0f
```
sprunk:
```suggestion
#define MAGIC_RETESSELATE_RETURNING_PATCH_CAMDIST (1000.0f * 1000.0f)
```

rts/Map/SMF/ROAM/Patch.h
```diff
@@ -24,11 +24,11 @@ class CCamera;
 
 // how many TriTreeNodes should be reserved per pool
 // (1M is a reasonable baseline for most large maps)
```
sprunk:
```suggestion
// (2M is a reasonable baseline for most large maps)
```

rts/Map/SMF/ROAM/RoamMeshDrawer.h
Beherith: So in my case, just use {} instead of {0} ?

rts/Map/SMF/ROAM/Patch.cpp
```diff
@@ -88,6 +90,7 @@ void CTriNodePool::Resize(size_t poolSize)
 	// live outside the pool, but KISS)
 	assert((poolSize & 1) == 0);
 	assert(poolSize > 0);
+	LOG_L(L_INFO, "[TriNodePool::%s] to %i ",__func__, poolSize);
```
sprunk: `poolSize` is `size_t` (ie. unsigned)
```suggestion
	LOG_L(L_INFO, "[TriNodePool::%s] to %u ",__func__, poolSize);
```

rts/Map/SMF/ROAM/RoamMeshDrawer.h
sprunk:
```suggestion
	std::array <int, MESH_COUNT> tesselationsSinceLastReset = {};
```

sprunk: yes

rts/Map/SMF/ROAM/Patch.cpp
ashdnazg: not `%u`, use `_STPF_` for printing `size_t`

sprunk: Possibly even `%zu` because `size_t` could be larger than a regular `unsigned` but I don't think `z` is portable.

ashdnazg: `%z` is not portable, that's why we use `_STPF_`

rts/Map/SMF/ROAM/Patch.cpp
ashdnazg: fine by me fwiw

rts/Map/SMF/ROAM/Patch.h
```diff
@@ -195,7 +207,7 @@ class Patch
 	// NOTE:
 	//   shadow-mesh patches are only ever viewed by one camera
 	//   normal-mesh patches can be viewed by *multiple* types!
-	std::array<unsigned int, CCamera::CAMTYPE_VISCUL> lastDrawFrames;
+	std::array<unsigned int, CCamera::CAMTYPE_VISCUL> lastDrawFrames = { 0 };
```
sprunk:
```suggestion
	std::array<unsigned int, CCamera::CAMTYPE_VISCUL> lastDrawFrames = {};
```

rts/Map/SMF/ROAM/RoamMeshDrawer.cpp
```diff
@@ -33,6 +33,11 @@ LOG_REGISTER_SECTION_GLOBAL(LOG_SECTION_ROAM)
 #define LOG_SECTION_CURRENT LOG_SECTION_ROAM
 
 
+//tessmode can be 1,2,(3 is LOAM)
+#define RETESSELLATE_MODE 3
+#define RETESSELATE_TO_FREE_INVISIBLE_PATCHES 2
```
sprunk: Inconsistent spelling (see the line above)
```suggestion
#define RETESSELLATE_TO_FREE_INVISIBLE_PATCHES 2
```

rts/Map/SMF/ROAM/RoamMeshDrawer.cpp
```diff
@@ -188,17 +195,81 @@ void CRoamMeshDrawer::Update()
...
 	bool tesselMesh = forceNextTesselation[shadowPass];
+	bool fullRetesselate = false;
 
 	auto& patches = patchMeshGrid[shadowPass];
 	auto& pvflags = patchVisFlags[shadowPass];
 
 	Patch::UpdateVisibility(cam, patches, numPatchesX);
 
-#define RETESSELLATE_MODE 1
+    int numPatchesVisible = 0;
```
sprunk: Maybe the various "number of things" could be `size_t`?
```suggestion
    size_t numPatchesVisible = 0;
```

Beherith:
all resolved, pulling anew

rts/Map/SMF/ROAM/RoamMeshDrawer.cpp
```diff
@@ -213,6 +218,7 @@ void CRoamMeshDrawer::Update()
 
 	CCamera* playerCamera = CCameraHandler::GetCamera(CCamera::CAMTYPE_PLAYER);
 	float3 playerCameraPosition = playerCamera->GetPos();
+	float totalCameraDistanceRatioInv = 0;
```
sprunk, line 221: type pedancy
```suggestion
	float totalCameraDistanceRatioInv = 0.0f;
```

rts/Map/SMF/ROAM/RoamMeshDrawer.cpp
```diff
@@ -341,36 +347,43 @@ void CRoamMeshDrawer::Update()
...
 					numPatchesEnterVisibility++;
-					if ((p.lastCameraPosition - lastCamPos[shadowPass]).SqLength() > MAGIC_RETESSELATE_RETURNING_PATCH_CAMDIST){
+
+
+				const float currentDistanceFromCamera = playerCameraPosition.distance(p.midPos);
+
+				//high numbers means we zoomed in more:
+				const float distanceFromCameraRatio = p.camDistanceLastTesselation  / currentDistanceFromCamera;
+
+				if((currentDistanceFromCamera > 500.0f ) && (distanceFromCameraRatio > MAGIC_RETESSELATE_CAMERA_RATIO)){
```
sprunk, line 359:
```suggestion
				if((currentDistanceFromCamera > MAGIC_RETESSELATE_CAMERA_DISTANCE) && (distanceFromCameraRatio > MAGIC_RETESSELATE_CAMERA_RATIO)){
```

rts/Map/SMF/ROAM/RoamMeshDrawer.cpp
```diff
@@ -473,9 +504,9 @@ void CRoamMeshDrawer::Update()
 		}
 	}
 	#if TESSMODE_3_DEBUG
-	if (actualTesselations> 0 || numPatchesEnterVisibility> 0 || numPatchesExitVisibility > 0 ){
+	if (actualTesselations> 0 ){
 
-		LOG("#Visible=%i delta_in=%i out=%i in df:#%i shadow:%i Z+%i actual_retess=%i up=%i skip=%i poolsize=%iK pooluse=%iK" ,
+		LOG("#V=%i d_in=%i d_out=%i in df:#%i shadow:%i Z+%i actual=%i up=%i skip=%i mratio = %.2f, P=%iK/%iK" ,
```
sprunk, line 509: some of those `"%i"` should be `_STFP_` due to `size_t`

ashdnazg:
thanks for the reference!

PR #508, opened 2020-03-31
LOAM (Lazy ROAM) for develop branch

Beherith:
ROAM frees up the entire triangle pool that stores the triangles of each tesselated patch (129*129 hmap pixels) and retesselates all patches whenever any in the visibility of patches occures, or if the unsynced heightmap changes (e.g.) explosion
LOAM does nothing when patches only go out of view, or when patches go into view, and a not-too-stale tesselation is aviable
when a 'stale' patch comes into view, only it is retesselated, and retesselation can affect neighbours, which is tracked and updated accordingly
when a heightmap change occurs, only that patch is retesselated, and if the neighbours are affected, that is updated accordingly
The cost: the mildly increased memory cost of keeping a larger triangle pool updated
the advantage is lag-free scrolling on a map
If the LOAM triangle pool runs out for any reason, then all the stored patches are all reset, and the whole view retesselated (exactly like it happens with ROAM on every visibility change)

### DIFF
--- a/rts/Map/Ground.cpp
+++ b/rts/Map/Ground.cpp
@@ -361,7 +361,7 @@ float CGround::LineGroundCol(float3 from, float3 to, bool synced)
 			// advance to the next nearest edge in either x or z dir, or in the case we reached the end make sure
 			// we set it to the exact square positions (floating point precision sometimes hinders us to hit it)
 			if (xn >= 1.0f && zn >= 1.0f) {
-				assert(curx != nextx || curz != nextz);
+				//assert(curx != nextx || curz != nextz);
 				curx = nextx;
 				curz = nextz;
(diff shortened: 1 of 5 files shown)


Beherith:
The conflict: 
			if (tesselMesh |= (uint8_t(p.IsVisible(cam)) != pvflags[i]))
				pvflags[i] = uint8_t(p.IsVisible(cam));
<<<<<<< develop
            numPatchesVisible+=pvflags[i];
=======
>>>>>>> develop
			if (p.IsVisible(cam) && p.IsDirty()) {
				p.ComputeVariance();

has no effect, either is ok

Beherith:
Ready for takeoff :)

ashdnazg:
Many thanks!

PR #509, opened 2020-04-02
Fix to #6376 (development version)

sanguinariojoe:
This is the best way (or the most robust at least) I found to fix the problem.

With these changes my widget may works either using...

* the archive name (e.g. 1944_Moro_River_V1)
* the file name (e.g. 1944_moro_river_v1.sd7)
* the local filepath (e.g. maps/1944_moro_river_v1.sd7)
* ~~the absolute filepath (e.g. /home/user/.spring/maps/1944_moro_river_v1.sd7)~~ (This was actually a widget trick)

**Tested just on maintenance version**

### DIFF
--- a/rts/Lua/LuaVFS.cpp
+++ b/rts/Lua/LuaVFS.cpp
@@ -373,40 +373,29 @@ int LuaVFS::UseArchive(lua_State* L)
 	if (CLuaHandle::GetHandleSynced(L))
 		return 0;
 
-	const std::string& filename = archiveScanner->ArchiveFromName(luaL_checkstring(L, 1));
-	const std::string& filepath = archiveScanner->GetArchivePath(filename) + filename;
-	const std::string& arname = archiveScanner->NameFromArchive(filename);
-
-	// FIXME: return 0, keep searches within the Spring directory
-	// the path may point to a file or dir outside of any data-dir
-	if (!LuaIO::IsSimplePath(filename)) {}
-
+	const std::string& archiveName = luaL_checkstring(L, 1);
+	const CArchiveScanner::ArchiveData& archiveData = archiveScanner->GetArchiveData(archiveName);
+	if (archiveData.IsEmpty())
+		return 0;
 
 	constexpr int funcIndex = 2;
-	int callError = 0;
-
 	if (!lua_isfunction(L, funcIndex))
 		return 0;
 
-	if (!CFileHandler::FileExists(filepath, SPRING_VFS_RAW))
-		return 0;
+	// block other threads from getting the global until we are done
+	vfsHandler->GrabLock();
+	vfsHandler->SetName("LuaVFS");
+	vfsHandler->UnMapArchives(false);
 
-	if (!filename.empty()) {
-		// block other threads from getting the global until we are done
-		vfsHandler->GrabLock();
-		vfsHandler->SetName("LuaVFS");
-		vfsHandler->UnMapArchives(false);
+	// could be mod,map,etc
+	vfsHandler->AddArchive(archiveName, false);
 
-		// could be mod,map,etc
-		vfsHandler->AddArchive(arname, false);
+	callError = lua_pcall(L, lua_gettop(L) - funcIndex, LUA_MULTRET, 0);
 
-		callError = lua_pcall(L, lua_gettop(L) - funcIndex, LUA_MULTRET, 0);
-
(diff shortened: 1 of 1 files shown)


abma:
i didn't investigate, but:

"the absolute filepath (e.g. /home/user/.spring/maps/1944_moro_river_v1.sd7)"

looks wrong as it doesn't sync i guess?!

sanguinariojoe:
Yeah, the point is that the widget (not `MapArchive()` either `UseArchive()`) is automagically shortening it to `maps/1944_moro_river_v1.sd7`.

I remove that from the description.

gajop:
I think the inputs are unnecessarily convoluted..
I also struggled using these functions before (MapArchive, UseArchive), and the LuaArchive:: API I created later is much easier to use, see https://github.com/spring/spring/blob/develop/rts/Lua/LuaArchive.cpp#L88 for example.

VFSHandler's AddArchive is expecting an archiveName, so I recommend we simplify things here. Make the first parameter simply be an archiveName?

Instead of doing these checks to the raw VFS
`if (!CFileHandler::FileExists(filename, SPRING_VFS_RAW)) {`
Simply check with VFSHandler's `HasArchive` (or use `AddArchiveIf`).

And do the same thing in MapArchive and UnmapArchive 

If you want to have a raw VFS filename/filepath -> archiveName conversion, that could be a separate function that Lua would have to use explicitly. The inverse of that already exists as LuaArchive::GetArchivePath I think, so if you want to add it, it would be under LuaArchive.

sanguinariojoe:
OK, now it only accepts archive names.

There is no need to add a new method to LUA archive, since you can always use (VFS.GetArchiveInfo): https://github.com/spring/spring/blob/9382c77df3d62e1ceb7cf6245d75d435e8ef4f4d/rts/Lua/LuaArchive.cpp#L118

Just for the record, there are some inconsistencies in regards the name considered by spring (which is apparently obtained from the smf file), and the name eventually defined in mapinfo.lua, considered by rapid. Something to be addressed in the future...

gajop:
Seems good, thanks!
This is a breaking change but apparently no one is using these functions anyway from what I can tell:
```
(base) ➜  pop-games grep -Rni VFS.MapArchive     
Chobby.sdd/LuaMenu/widgets/zk_campaign_handler.lua:420:--		--VFS.MapArchive(missionArchive)
(base) ➜  pop-games grep -Rni VFS.UseArchive
(base) ➜  pop-games ls
Balanced-Annihilation.sdd  BYAR.sdd  Chobby.sdd  EvoRTS.sdd  JauriaRTS.sdd  spring1944.sdd  SpringMCL.sdd  Zero-K.sdd
(base) ➜  pop-games 
```

sanguinariojoe:
I am! I am creating an in-game lobby (something different to chobby):

https://github.com/sanguinariojoe/s44lobby

gajop:
> I am! I am creating an in-game lobby (something different to chobby):
> 
> https://github.com/sanguinariojoe/s44lobby
> 
> Thanks BTW!

https://github.com/sanguinariojoe/s44lobby/blob/c35d36c0fd8af548c5a3266fead703965e4c1559/LuaUI/Widgets/cmd_getminimap.lua

absolute madman :) 

I was looking at how engine is providing minimap/infomap loading functionality to unitsync. Having MapArchive/UseArchive doesn't really help us that much, as we still have to do the same painful loading process you're doing there.

I think engine should be helping there. It already knows how to load maps, so maybe we should just expose that functionality?
Not yet sure what the API should be like... Perhaps it could load map objects (Lua userdata) that we could query textures from. Or it provide a per-texture query API, similar to unitsync

sanguinariojoe:
> absolute madman :)

A bit :-p

> I was looking at how engine is providing minimap/infomap loading functionality to unitsync. Having MapArchive/UseArchive doesn't really help us that much, as we still have to do the same painful loading process you're doing there.
> 
> I think engine should be helping there. It already knows how to load maps, so maybe we should just expose that functionality?
> Not yet sure what the API should be like... Perhaps it could load map objects (Lua userdata) that we could query textures from. Or it provide a per-texture query API, similar to unitsync

I was actually thinking on doing that, but I though changing the API so much is not the definition of maintenance. Maybe I'm coming back to this topic when jumping to develop.

Along this line, cherry pick this for maintenance branch please

gajop:
> Along this line, cherry pick this for maintenance branch please

I have already

PR #511, opened 2020-04-05
Coerce to number less often in *SetRulesParam

esainane:
Unlike most lua_is* functions, lua_isnumber checks to see if the
given value *can* be a number, rather than if it currently is one:

> Returns 1 if the value at the given acceptable index is a number or
> a string convertible to a number, and 0 otherwise.

This means that even if a lua caller expressly coerces to string
beforehand, as long as the string looks like a number, it gets stored
as a number. This can cause problems if, for example, a player name
that looks like a very large number loses precision.

See esainane/zkstats@99ddb35 for an example.

Explicitly checking using the lua_israwnumber variant allows the caller
to provide either a string or a number, and have it be stored as such.

`develop` variant of #503.


ashdnazg:
TY!

PR #515, opened 2020-04-09
 LOAM: early bailout for visibility calculations on no change, reduce…

Beherith:
… for_mt overhead

A. There are 4 conditions that can require a visibility/retess check:
1. Camera position change
2. camera direction change
3. heightmap change
4. grounddetail param change
Otherwise we can bail on a lot of expensive things

B. for_mt seems to have a high setup cost, and spins threads needlessly.
for_mt vertex index generation is only useful when a full retesselation is performed, otherwise the number of patches that need their indices regenerated is likely less than the number of threads.

C. side note: /debugview performance is now not strongly affected, side effect of less for_mt?


Beherith:
Note threadpool load before:
screen00032

Threadpool load after:
screen00035

ashdnazg:
ty

PR #516, opened 2020-04-12
Fix non-visible enemy units and structures not blocking construction

GoogleFrog:
This PR does two things:
- Fix a common structure stacking bug noticed since 104.0.1-1477-g8ecf38a was adopted by ZK.
- Fix a case related to https://springrts.com/mantis/view.php?id=6346

Currently it is possible to start construction on a location blocked by an enemy structure provided that the enemy structure is not within LOS. This can be done with luaui disabled and I expect that it could be done with luarules disabled as well.

image

The cause is the fix for https://springrts.com/mantis/view.php?id=6346 by https://github.com/spring/spring/commit/7cd04e709d52cec18231cff4413f8ddaefb5adb7.

Prior to https://github.com/spring/spring/commit/7cd04e709d52cec18231cff4413f8ddaefb5adb7 'u' could be any unit that was blocking the square. If 'u' could not be assisted there was a buildInfo.FootPrintOverlap check that caused StartBuild to return false, blocking construction. It seems like this was the only check stopping units from starting constructions on unseen enemy units, and this makes sense because blocking the construction earlier (such as when the order is given) would open the door for maphax.

After https://github.com/spring/spring/commit/7cd04e709d52cec18231cff4413f8ddaefb5adb7 the only values that the variable 'u' can take are nullptr or that of an assistable unit or of a friendly unit. The buildInfo.FootPrintOverlap is never reached for enemy units so StartBuild cannot return false in the CGameHelper::BUILDSQUARE_OCCUPIED case if enemy units are the only blockers.

While checking whether this PR caused a regression of https://springrts.com/mantis/view.php?id=6346 I found a case of structure stacking that had not been fixed. It can be reproduced on South-East corners on Fifteen Platforms.

image

The code added to the CGameHelper::BUILDSQUARE_OPEN case prevents the previously unfixed case of structure stacking.

### DIFF
--- a/rts/Sim/Units/UnitTypes/Builder.cpp
+++ b/rts/Sim/Units/UnitTypes/Builder.cpp
@@ -748,7 +748,7 @@ bool CBuilder::StartBuild(BuildInfo& buildInfo, CFeature*& feature, bool& inWait
 
 				if (cu == nullptr)
 					continue;
-				if (allyteam != cu->allyteam)
+				if (!teamHandler.Ally(allyteam, cu->allyteam)
 					return false; // Enemy units that block always block the cell
 				if (!CanAssistUnit(cu, buildInfo.def))
 					continue;
--- a/rts/Sim/Units/UnitTypes/Builder.cpp
+++ b/rts/Sim/Units/UnitTypes/Builder.cpp
@@ -708,8 +708,25 @@ bool CBuilder::StartBuild(BuildInfo& buildInfo, CFeature*& feature, bool& inWait
 	const CGameHelper::BuildSquareStatus tbs = CGameHelper::TestUnitBuildSquare(buildInfo, feature, -1, true);
 
 	switch (tbs) {
-		case CGameHelper::BUILDSQUARE_OPEN:
-			break;
+		case CGameHelper::BUILDSQUARE_OPEN: {
+			// <pos> might map to a non-blocking portion
+			// of the buildee's yardmap, fallback check
+			const CUnit* u = CGameHelper::GetClosestFriendlyUnit(nullptr, buildInfo.pos, 8.0f, allyteam);
+			if (u != nullptr) {
+				if (CanAssistUnit(u, buildInfo.def)) {
+					// StopBuild sets this to false, fix it here if picking up the same buildee again
+					terraforming = (u == prvBuild && u->terraformLeft > 0.0f);
+
+					AddDeathDependence(curBuild = const_cast<CUnit*>(u), DEPENDENCE_BUILD);
+					ScriptStartBuilding(u->pos, false);
+					return true;
+				}
+
+				// let BuggerOff handle this case (TODO: non-landed aircraft should not count)
+				if (buildInfo.FootPrintOverlap(u->pos, u->GetFootPrint(SQUARE_SIZE * 0.5f)))
+					return false;
+			}
+		} break;
 
 		case CGameHelper::BUILDSQUARE_BLOCKED:
 		case CGameHelper::BUILDSQUARE_OCCUPIED: {
@@ -731,11 +748,12 @@ bool CBuilder::StartBuild(BuildInfo& buildInfo, CFeature*& feature, bool& inWait
 
 				if (cu == nullptr)
 					continue;
+				if (allyteam != cu->allyteam)
+					return false; // Enemy units that block always block the cell
 				if (!CanAssistUnit(cu, buildInfo.def))
 					continue;
 
 				u = cu;
-				break;
 			}
 
 			// <pos> might map to a non-blocking portion
--- a/rts/Sim/Units/UnitTypes/Builder.cpp
+++ b/rts/Sim/Units/UnitTypes/Builder.cpp
@@ -711,7 +711,7 @@ bool CBuilder::StartBuild(BuildInfo& buildInfo, CFeature*& feature, bool& inWait
 		case CGameHelper::BUILDSQUARE_OPEN: {
 			// <pos> might map to a non-blocking portion
 			// of the buildee's yardmap, fallback check
-			const CUnit* u = CGameHelper::GetClosestFriendlyUnit(nullptr, buildInfo.pos, 8.0f, allyteam);
+			const CUnit* u = CGameHelper::GetClosestFriendlyUnit(nullptr, buildInfo.pos, static_cast<float>(SQUARE_SIZE), allyteam);
 			if (u != nullptr) {
 				if (CanAssistUnit(u, buildInfo.def)) {
 					// StopBuild sets this to false, fix it here if picking up the same buildee again


rts/Sim/Units/UnitTypes/Builder.cpp
```diff
@@ -708,8 +708,25 @@ bool CBuilder::StartBuild(BuildInfo& buildInfo, CFeature*& feature, bool& inWait
 	const CGameHelper::BuildSquareStatus tbs = CGameHelper::TestUnitBuildSquare(buildInfo, feature, -1, true);
 
 	switch (tbs) {
-		case CGameHelper::BUILDSQUARE_OPEN:
-			break;
+		case CGameHelper::BUILDSQUARE_OPEN: {
+			// <pos> might map to a non-blocking portion
+			// of the buildee's yardmap, fallback check
+			const CUnit* u = CGameHelper::GetClosestFriendlyUnit(nullptr, buildInfo.pos, 8.0f, allyteam);
```
sprunk, line 714: I see yardmap referred to in the comment above, is the magic  `8.0f` actually `SQUARE_SIZE`?
```suggestion
			const CUnit* u = CGameHelper::GetClosestFriendlyUnit(nullptr, buildInfo.pos, static_cast<float>(SQUARE_SIZE), allyteam);
```

rts/Sim/Units/UnitTypes/Builder.cpp
```diff
@@ -731,11 +748,12 @@ bool CBuilder::StartBuild(BuildInfo& buildInfo, CFeature*& feature, bool& inWait
 
 				if (cu == nullptr)
 					continue;
+				if (allyteam != cu->allyteam)
```
ashdnazg, line 751:
```suggestion
				if (!teamHandler.Ally(allyteam, cu->allyteam)
```

rts/Sim/Units/UnitTypes/Builder.cpp
```diff
@@ -708,8 +708,25 @@ bool CBuilder::StartBuild(BuildInfo& buildInfo, CFeature*& feature, bool& inWait
...
+					AddDeathDependence(curBuild = const_cast<CUnit*>(u), DEPENDENCE_BUILD);
+					ScriptStartBuilding(u->pos, false);
+					return true;
+				}
+
+				// let BuggerOff handle this case (TODO: non-landed aircraft should not count)
+				if (buildInfo.FootPrintOverlap(u->pos, u->GetFootPrint(SQUARE_SIZE * 0.5f)))
+					return false;
+			}
+		} break;
```
ashdnazg, line 729: Why is there BUILDSQUARE_OPEN returned in the first place?

rts/Sim/Units/UnitTypes/Builder.cpp
```diff
@@ -708,8 +708,25 @@ bool CBuilder::StartBuild(BuildInfo& buildInfo, CFeature*& feature, bool& inWait
...
+
+					AddDeathDependence(curBuild = const_cast<CUnit*>(u), DEPENDENCE_BUILD);
+					ScriptStartBuilding(u->pos, false);
+					return true;
+				}
+
+				// let BuggerOff handle this case (TODO: non-landed aircraft should not count)
+				if (buildInfo.FootPrintOverlap(u->pos, u->GetFootPrint(SQUARE_SIZE * 0.5f)))
+					return false;
+			}
```
sprunk, line 728: Could do an early return:
```suggestion
			if (u == nullptr)
				break;

			if (CanAssistUnit(u, buildInfo.def)) {
				// StopBuild sets this to false, fix it here if picking up the same buildee again
				terraforming = (u == prvBuild && u->terraformLeft > 0.0f);

				AddDeathDependence(curBuild = const_cast<CUnit*>(u), DEPENDENCE_BUILD);
				ScriptStartBuilding(u->pos, false);
				return true;
			}

			// let BuggerOff handle this case (TODO: non-landed aircraft should not count)
			if (buildInfo.FootPrintOverlap(u->pos, u->GetFootPrint(SQUARE_SIZE * 0.5f)))
				return false;
			break;
```

rts/Sim/Units/UnitTypes/Builder.cpp
```diff
@@ -731,11 +748,12 @@ bool CBuilder::StartBuild(BuildInfo& buildInfo, CFeature*& feature, bool& inWait
 
 				if (cu == nullptr)
 					continue;
+				if (allyteam != cu->allyteam)
+					return false; // Enemy units that block always block the cell
 				if (!CanAssistUnit(cu, buildInfo.def))
 					continue;
 
 				u = cu;
-				break;
```
sprunk, line 738: Since the loop now continues, `o` (see line 730/747) can refer to something other than `u` after the loop (for example if there is one `nullptr` in `cells` afterwards). On the other hand I don't see `o` being used so it might be a good moment to remove it.

rts/Sim/Units/UnitTypes/Builder.cpp
sprunk: If you're allied in an FFA you still don't share LoS and potentially still can't see the unit though?

rts/Sim/Units/UnitTypes/Builder.cpp
sprunk: See the comment: `// <pos> might map to a non-blocking portion of the buildee's yardmap`

sprunk: Though that comment seems incorrect since ZK windgens don't have a non-blocking portion of the yardmap?

https://github.com/ZeroK-RTS/Zero-K/blob/40e6defe2425b46518cd37105c2b6380db4f920b/units/energywind.lua#L46

ashdnazg: no, this comment was copied as is from below

rts/Sim/Units/UnitTypes/Builder.cpp
ashdnazg: you can assist them, though, AFAIK.

sprunk: You can assist enemies too though, I've seen automexing cons assisting enemy mexes

rts/Sim/Units/UnitTypes/Builder.cpp
GoogleFrog: No but it can be.

rts/Sim/Units/UnitTypes/Builder.cpp
GoogleFrog: > Why is there BUILDSQUARE_OPEN returned in the first place?

Basically because CGameHelper::TestUnitBuildSquare is wrong in rare but reproducible cases. See the second image in the OP. You can remove this whole change if you want.

ashdnazg: Have you tried to debug CGameHelper::TestUnitBuildSquare and fix it?

rts/Sim/Units/UnitTypes/Builder.cpp
```diff
@@ -731,11 +748,12 @@ bool CBuilder::StartBuild(BuildInfo& buildInfo, CFeature*& feature, bool& inWait
 
 				if (cu == nullptr)
 					continue;
+				if (!teamHandler.Ally(allyteam, cu->allyteam)
```
sprunk, line 751: I have doubts. Here's the callouts to put in some gadget that ceasefires both teams for a test (the rest of the test setup stays the same as previous since ZK doesn't have allied victory AFAIK and LoS isn't shared):
```lua
Spring.SetAlly(0, 1)
Spring.SetAlly(1, 0)
```

ashdnazg: doubts about what?

GoogleFrog: Doubts that checking whether the teams are allied is correct to do here.

GoogleFrog: Anyway, the suggested change seems to be missing a parenthesis.

GoogleFrog:
The PR has become broken by me clicking the "use suggestion" button and I don't care enough to fight all of it at this point in time. See https://github.com/spring/spring/pull/518.

PR #517, opened 2020-04-12
Fix roam maintenance

Beherith:


rts/Map/SMF/ROAM/RoamMeshDrawer.cpp
```diff
@@ -398,6 +398,8 @@ void CRoamMeshDrawer::Update()
 			});
 		} else {
 			for (Patch& p: patches) {
+				if (!p.isChanged)
```
ashdnazg, line 401: shouldn't this be added in the `for_mt` version as well?

ashdnazg: can be copied straight from here:
https://github.com/spring/spring/blob/develop/rts/Map/SMF/ROAM/RoamMeshDrawer.cpp#L389

Beherith: for_mt is only used when all visible patches are retesselated, and if they are retesselated, they are changed.
The reason for this is not to spin up a bunch of threads needlessly, since usually only 1-2 patches are retesselated.

ashdnazg: that means the develop version doesn't need it.

Beherith: The develop version needs it too. 
There is a bit of confusion regarding the MT in ROAM. The problematic MT was the tesselation of multiple patches in threads, with multple tripools, this is what your previous commit ripped out. 
The safe MT in ROAM is the generation of indices, which is actually beneficial when the entire thing is reset & retesselated (due to staleness, detail change, high level zoom change).

Beherith: To make things clear and concise, they can be identical in develop and maintenance.

ashdnazg: This is what I meant

ashdnazg:
TY, added to both develop and maintenance.

PR #519, opened 2020-04-13
Minor cleanup

sprunk:


rts/Sim/Units/UnitTypes/Builder.cpp
```diff
@@ -722,16 +722,15 @@ bool CBuilder::StartBuild(BuildInfo& buildInfo, CFeature*& feature, bool& inWait
 			// closest to us
 			const CGroundBlockingObjectMap::BlockingMapCell& cell = groundBlockingObjectMap.GetCellUnsafeConst(buildInfo.pos);
 
-			const CSolidObject* o = nullptr;
 			const CUnit* u = nullptr;
 
 			// look for any blocking assistable buildee at build.pos
 			for (size_t i = 0, n = cell.size(); i < n; i++) {
-				const CUnit* cu = dynamic_cast<const CUnit*>(o = cell[i]);
+				const CUnit* cu = dynamic_cast<const CUnit*>(cell[i]);
```
gajop, line 729: why even const cast?

rts/Sim/Units/UnitTypes/Builder.cpp
```diff
@@ -722,16 +722,15 @@ bool CBuilder::StartBuild(BuildInfo& buildInfo, CFeature*& feature, bool& inWait
...
 
 				if (cu == nullptr)
 					continue;
-				if (allyteam != cu->allyteam)
+				if (allyteam != cu->allyteam) // NB: not teamHandler.Ally because allies don't share LoS
```
ashdnazg, line 733: I don't think LoS is relevant here.
The question is whether the default action for trying to build a building on the same spot an ally is building should be to assist.

rts/Sim/Units/UnitTypes/Builder.cpp
sprunk: `const` has to be in the cast because `cu` is `const *`, and `cu` being `const *` is probably a decent idea because we don't want to modify it. It's type repetition though, how about
```suggestion
				auto cu = dynamic_cast<const CUnit*>(cell[i]);
```

rts/Sim/Units/UnitTypes/Builder.cpp
sprunk: I think whether to default to assisting for allies is a valid but orthogonal question because the new stacking bug was about starting construction on top of finished buildings (which are unassistable). I now agree LoS isn't relevant, the real reason `teamHandler.Ally` can't be used is that `GetClosestFriendlyUnit` (which alongside the subsequent `buildInfo.FootPrintOverlap` are the fallback preventing stacking on finished, ie. unassistable buildings) works only on the allyteam and not its allies. Personally I don't think allies should be assisted by default (since they are really just temporarily ceasefired enemies) but I am probably not the one to decide here and I mostly just wanted to spend 5 seconds writing a comment preventing people with good intentions from reintroducing the stacking bug.
```suggestion
				/* NB: allies (cf. `teamHandler.Ally`) are also blocked
				 * even though in theory we could want to assist them. This
				 * is because allies aren't caught by `GetClosestFriendlyUnit`
				 * so it would be possible to build on top of them if they
				 * weren't assistable. */
				if (allyteam != cu->allyteam)
```

gajop:
Don't know about the comment but I approve the code change.

rts/Sim/Units/UnitTypes/Builder.cpp
```diff
@@ -722,15 +722,19 @@ bool CBuilder::StartBuild(BuildInfo& buildInfo, CFeature*& feature, bool& inWait
...
-				const CUnit* cu = dynamic_cast<const CUnit*>(o = cell[i]);
+				const CUnit* cu = dynamic_cast<const CUnit*>(cell[i]);
 
 				if (cu == nullptr)
 					continue;
+				/* NB: allies (cf. `teamHandler.Ally`) are also blocked
+				 * even though in theory we could want to assist them. This
+				 * is because allies aren't caught by `GetClosestFriendlyUnit`
+				 * so it would be possible to build on top of them if they
+				 * weren't assistable. */
```
gajop, line 737: Can you further elaborate on this comment?

I've looked in the code a bit and it seems that this invokes
https://github.com/spring/spring/blob/develop/rts/Game/GameHelper.cpp#L801-L806
which in turn uses the Friendly class to filter units. The per-team filter there uses the function you mentioned https://github.com/spring/spring/blob/develop/rts/Game/GameHelper.cpp#L406 , so I'm not sure how they "aren't getting caught"?

sprunk: Now I'm unsure. Maybe there was some issue with the tests. I think it would be the best if I just remove the comment and maybe come up with a better one later if the situation clears up.

PR #521, opened 2020-04-19
AllowSpectatorJoin defaults to 0 for dedi

sprunk:
According to @abma this setting should ideally be disabled for autohosts, https://github.com/ZeroK-RTS/Zero-K-Infrastructure/issues/789#issuecomment-139229098


gajop:
Seems reasonable, thanks

PR #522, opened 2020-04-23
Avoiding "push resistance" stucked units (develop version)

sanguinariojoe:

https://youtu.be/0ZxID60eJaU

(Tested just on maintenance)


rts/Sim/MoveTypes/GroundMoveType.cpp
```diff
@@ -2101,6 +2101,20 @@ void CGroundMoveType::HandleUnitCollisions(
 			const bool allowNewPath = (!atEndOfPath && !atGoal);
 			const bool checkYardMap = ((pushCollider || pushCollidee) || collideeUD->IsFactoryUnit());
 
+			// NOTE:
+			//   This is superhacky... But works like a charm!!!
+			//   In case 2 push resistant units park together, they would not
+			//   separate anymore, since pathfinding is not able to find a
+			//   non-blocked route.
+			//   This may happens for instance in unit balls, where units can
+			//   be pushed twiced, coming back to the blocked position.
```
gajop, line 2110: twiced -> twice!

sanguinariojoe:
> Is this an urgent issue for you?

We are having many problems with blocked units in s:44. However, I cannot say if it a good risk-benefit balance, since I don't know how many other games are using push resistance.

sanguinariojoe:
Hey!

We (Spring-1944) can really enjoy these changes. Can someone please take a look, merge, and cherry-pick in maintenance?

I have been testing for a while, and I think the solution is robust and efficient.

ashdnazg:
needs bumping path estimator version

abma:
am i right with:

push resistence basicly is a hack to simulate small units can't push large units?

if so, shouldn't push resistence beeing removed and replaced by more realistic physics which takes mass and friction of objects into account?


can't stuck units beeing detected by the timeout of move commands and then moved by a lua gadged?

please correct me if i'm wrong! thank you!

sanguinariojoe:
> push resistence basicly is a hack to simulate small units can't push large units?

Nope. Push resistance completely disables the units pushing system, no matter the mass/size of the units. And actually that is the behavior we want in S44

> can't stuck units beeing detected by the timeout of move commands and then moved by a lua gadged?

I tried this approach in the past with quite poor results. I really think this approach is the appropriate one, i.e. if 2 units are trying to get the same spot, just one of them may take it and the other one shall "keep moving"

PR #523, opened 2020-04-23
Avoiding "push resistance" stucked units (maintenance version)

sanguinariojoe:

https://youtu.be/0ZxID60eJaU

(Tested just on maintenance)


sanguinariojoe:
I always mess up with the branches...

PR #524, opened 2020-04-23
Avoiding "push resistance" stucked units (maintenance version)

sanguinariojoe:

https://youtu.be/0ZxID60eJaU

(Tested just on maintenance)


GoogleFrog:
Your footprints are non-square? That is terrifying and I'll be impressed if it works. I might look into using pushResistant again if this works.

sanguinariojoe:
> Your footprints are non-square?

Nope, see `modInfo.allowSepAxisCollisionTest`

> That is terrifying and I'll be impressed if it works.

It is indeed. Unfortunately a more smart solution would require diving into blocking maps, hacking much more other parts of the code.
Phylosophically, it is saying something like "you are allowed to stop here for the time being, but you cannot stand here because this spot is mine''

gajop:
Once (if) this gets merged to develop (https://github.com/spring/spring/pull/522), we'll manually cherry pick it to maintenance. No need to make a separate PR.

PR #525, opened 2020-04-27
Spring.SetUnitNoFoo: expose to LuaUI

sprunk:
Games which don't want those functions used can just do `Spring.SetUnitNoFoo = nil` at LuaUI entry point.


ashdnazg:
IIRC the luaui entry point is overridable from outside the game archive.

sprunk:
Nope, see c412571058a9a2e6cf7fae237f26f66b53644412

gajop:
I think you should also get rid of GetHandleUserMode since it's not used anymore.
GetUserMode and userMode may have to stay because of the CLuaHandle::Save call-in. 

Otherwise you have a +1 from me. It's good that unsynced widgets and unsynced gadgets match on what they can do through the LuaUnsyncedCtrl interface.

PR #528, opened 2020-05-03
Cannon: Use autoTargetRangeBoost when set

esainane:
Resolves mantis #6382: https://springrts.com/mantis/view.php?id=6382

### DIFF
--- a/doc/changelog.txt
+++ b/doc/changelog.txt
@@ -186,6 +186,9 @@ Lua:
      spring -> rx, ry, rz, dist
      free -> rx, ry, rz, vx, vy, vz, avx, avy, avz
      ov -> (nil)
+ - Spring.SetUnitWeaponState with "autoTargetRangeBoost" now also lets Cannon
+   and StarburstLauncher weapons look ahead and pre-aim at targets just
+   outside of nominal range.
 
 AI:
  - reveal unit's captureProgress, buildProgress and paralyzeDamage params through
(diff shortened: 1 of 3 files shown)


esainane:
I'm not sure what the best way to create a sister PR for the maintenance branch is. Should I make a merge commit, or just cherry-pick it?

ashdnazg:
don't make sister PR, this will be cherry-picked after merging

gajop:
For reference, this is the original commit that introduced the unused boost variables (perhaps on purpose): https://github.com/spring/spring/commit/15d6cc6661b00ebd5650308b3d723e9ae3a21047

If you add it to Cannon, you may as well add it to StarburstLauncher?

Other than that, seeing as 0 is the default I think this could be merged. 
So you have a very weak accept from me if you apply the same thing to the StartburstLauncher

esainane:
Added!

gajop:
@esainane Ah.. I forgot to ask, but can you update the changelog too please?

esainane:
Would something like the following in the lua section of doc/changelog.txt be appropriate?

```
 - Spring.SetUnitWeaponState with "autoTargetRangeBoost" now also lets Cannon
   and StarburstLauncher weapons look ahead and pre-aim at targets just
   outside of nominal range.
```

gajop:
I suppose that's fine

esainane:
Added it in. Any suggestions if the wording could be improved would be more than welcome, of course!

PR #530, opened 2020-05-13
Fix floating units leaving tracks on sea floor

a-nop:
This allows defining units which both leave tracks on ground and float on water, such as bots that swim on the surface.

### DIFF
--- a/rts/Rendering/Env/Decals/LegacyTrackHandler.cpp
+++ b/rts/Rendering/Env/Decals/LegacyTrackHandler.cpp
@@ -421,6 +421,8 @@ void LegacyTrackHandler::AddTrack(const CUnit* unit, const float3& newPos)
 		return;
 	if (!unitDef->IsGroundUnit())
 		return;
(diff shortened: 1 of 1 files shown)


a-nop:
It looks like an apt hiccup broke the CI job.

rts/Rendering/Env/Decals/LegacyTrackHandler.cpp
```diff
@@ -421,6 +421,8 @@ void LegacyTrackHandler::AddTrack(const CUnit* unit, const float3& newPos)
 		return;
 	if (!unitDef->IsGroundUnit())
 		return;
+	if (unitDef->floatOnWater && newPos[1] <= unitDef->waterline * -1.0f)
```
rtri, line 424: This can be reduced to `if (unit->IsInWater() && !unit->IsOnGround())`, a unit's waterline value might change after it is loaded.

a-nop: Thanks @rtri, that's much better.

PR #532, opened 2020-06-01
[WIP] GetUnitLastTracedPiece

sprunk:
Same as Spring.GetUnitLastHitPiece, but for unsynced (ray traces etc)

### DIFF
--- a/doc/changelog.txt
+++ b/doc/changelog.txt
@@ -18,6 +18,9 @@ Sim:
 
 Lua:
  - add math.tau
+ - add Spring.Get{Unit,Feature}LastTracedPiece
+   Same as Spring.Get{Unit,Feature}LastHitPiece,
+   but for unsynced (ray traces etc)
  - add sensors.decloakRequiresLineOfSight modrule. If true, sight radius works as a cap for decloak radius
    (for example a unit with decloak radius 200 will not be decloaked with a 100 LoS unit in the 100-200 range,
    and units with 0 LoS can't decloak others at all regardless of decloak range). Defaults to false.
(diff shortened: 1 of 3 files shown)


lhog:
@sprunk Can you describe the use case behind this proposal, please?

sprunk:
IIRC Flozi wanted to be able to highlight individual pieces by mousing over them in S44, because individual pieces matter there (can get damaged individually, as facilitated by the synced `Spring.GetUnitLastAttackedPiece`, so you might want to check their HP as well - this is supposed to just be an unsynced counterpart). I agreed that it would be good to have something that lets you read the piece under a mouse raytrace so I made this as a blind proof-of-concept patch so that Flozi (or whoever it was) would be able to find it and finish it (apparently this didn't happen).

This probably isn't even the best way to implement that use case (`Spring.TraceRay` could probably just return `pieceID` in addition to `{unit,feature}ID`) and I don't have much of a personal stake here so I don't mind this getting closed.

lhog:
@FLOZi what's your take on this?

sprunk:
I'm not fully certain if it was actually Flozi.

lhog:
BTW, are you sure this code does what expected? I haven't checked on the develop, but on transition I found no evidences `hitModelPieces[false]` has anything in common with ray tracing.

sprunk:
You're right. I'm fairly sure it was _somehow_ meant to be related to finding the piece under the mouse but my memory is failing me and it seems clear that this isn't the way to implement that use case. I don't think this PR will be of any use to anyone.

sprunk:
For the record Flozi says it was likely him but isn't bothered too much about the intended functionality not yet existing.

PR #535, opened 2020-06-03
Fixed particles clipping into existing geometry (so-called soft edge …

lhog:
This one implements "soft edges" of the particles, similar to how all other engines (Unity, UE4, Godot) do that.

This time around it's only for maintenance. I have no game to test it on develop, but should be easy to integrate it in the future.

No soft edges, status quo, visible clipping areas
screen01784

Softened edges (5.0 elmo, hardcoded), no visible clipping areas
screen01785

### DIFF
--- a/cont/base/springcontent/shaders/GLSL/ProjFXFragProg.glsl
+++ b/cont/base/springcontent/shaders/GLSL/ProjFXFragProg.glsl
@@ -8,26 +8,28 @@ in vec4 vsPos;
 
 #define projMatrix gl_ProjectionMatrix
 
-uniform vec2 invScreenRes;
+uniform vec2 invScreenSize;
 
 #define NORM2SNORM(value) (value * 2.0 - 1.0)
 #define SNORM2NORM(value) (value * 0.5 + 0.5)
 
-float GetViewSpaceDepth(float depthNDC) {
-	return -projMatrix[3][2] / (projMatrix[2][2] + depthNDC);
-}
-
-float GetViewSpaceDepthFromTexture(float depthNDC) {
+float GetViewSpaceDepth(float d) {
 	#ifndef DEPTH_CLIP01
-		depthNDC = NORM2SNORM(depthNDC);
+		d = NORM2SNORM(d);
 	#endif
-	return GetViewSpaceDepth(depthNDC);
+	return -projMatrix[3][2] / (projMatrix[2][2] + d);
 }
 
 void main() {
 	vec4 color = texture(atlasTex, gl_TexCoord[0].xy);
-	float depthVS = GetViewSpaceDepthFromTexture(texture(depthTex, gl_FragCoord.xy * invScreenRes).x);
-	float edgeSmoothness = smoothstep(0.0, 5.0, vsPos.z - depthVS);
+	vec2 screenUV = gl_FragCoord.xy * invScreenSize;
+
+	float depthZO = texture(depthTex, screenUV).x;
+	float depthVS = GetViewSpaceDepth(depthZO);
+
+	float edgeSmoothness = smoothstep(-16.0, 0.0, vsPos.z - depthVS);
+
 	gl_FragColor = color * vertColor;
 	gl_FragColor *= edgeSmoothness;
+	gl_FragDepth = min(gl_FragCoord.z, depthZO - 1e-3); //always 0 to -1, no matter clip_ctrl
 }
--- a/cont/base/springcontent/shaders/GLSL/ProjFXFragProg.glsl
+++ b/cont/base/springcontent/shaders/GLSL/ProjFXFragProg.glsl
@@ -1,0 +1,33 @@
+#version 130
+
+uniform sampler2D atlasTex;
+uniform sampler2D depthTex;
+
+in vec4 vertColor;
+in vec4 vsPos;
+
+#define projMatrix gl_ProjectionMatrix
+
+uniform vec2 invScreenRes;
+
+#define NORM2SNORM(value) (value * 2.0 - 1.0)
+#define SNORM2NORM(value) (value * 0.5 + 0.5)
+
+float GetViewSpaceDepth(float depthNDC) {
+	return -projMatrix[3][2] / (projMatrix[2][2] + depthNDC);
+}
+
+float GetViewSpaceDepthFromTexture(float depthNDC) {
+	#ifndef DEPTH_CLIP01
+		depthNDC = NORM2SNORM(depthNDC);
+	#endif
+	return GetViewSpaceDepth(depthNDC);
+}
+
+void main() {
+	vec4 color = texture(atlasTex, gl_TexCoord[0].xy);
+	float depthVS = GetViewSpaceDepthFromTexture(texture(depthTex, gl_FragCoord.xy * invScreenRes).x);
+	float edgeSmoothness = smoothstep(0.0, 5.0, vsPos.z - depthVS);
+	gl_FragColor = color * vertColor;
+	gl_FragColor *= edgeSmoothness;
+}
--- a/cont/base/springcontent/shaders/GLSL/ProjFXVertProg.glsl
+++ b/cont/base/springcontent/shaders/GLSL/ProjFXVertProg.glsl
@@ -1,0 +1,11 @@
+#version 130
+
+out vec4 vertColor;
+out vec4 vsPos;
+
+void main() {
+	gl_TexCoord[0] = gl_MultiTexCoord0;
+	vertColor = gl_Color;
+	vsPos = gl_ModelViewMatrix * gl_Vertex;
+    gl_Position = gl_ProjectionMatrix * vsPos;
+}
--- a/cont/base/springcontent/shaders/GLSL/ProjFXVertProg.glsl
+++ b/cont/base/springcontent/shaders/GLSL/ProjFXVertProg.glsl
@@ -7,5 +7,5 @@ void main() {
 	gl_TexCoord[0] = gl_MultiTexCoord0;
 	vertColor = gl_Color;
 	vsPos = gl_ModelViewMatrix * gl_Vertex;
-    gl_Position = gl_ProjectionMatrix * vsPos;
+	gl_Position = gl_ProjectionMatrix * vsPos;
 }
--- a/rts/Rendering/Env/Particles/ProjectileDrawer.cpp
+++ b/rts/Rendering/Env/Particles/ProjectileDrawer.cpp
@@ -279,6 +279,7 @@ void CProjectileDrawer::Init() {
 			fxShader = shaderHandler->CreateProgramObject("[ProjectileDrawer::VFS]", "FX Shader", false);
 			fxShader->AttachShaderObject(shaderHandler->CreateShaderObject("GLSL/ProjFXVertProg.glsl", "", GL_VERTEX_SHADER));
 			fxShader->AttachShaderObject(shaderHandler->CreateShaderObject("GLSL/ProjFXFragProg.glsl", "", GL_FRAGMENT_SHADER));
+			fxShader->SetFlag("DEPTH_CLIP01", globalRendering->supportClipSpaceControl);
 			fxShader->Enable();
 			fxShader->SetUniform("atlasTex", 0);
 			fxShader->SetUniform("depthTex", 8);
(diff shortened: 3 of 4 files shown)


lhog:
I'm likely be able to fix ground flashes too in the following commits

rtri:
Please stop adding graphics code to maintenance.

lhog:
I can stop, but doing so won't move the needle on migration to develop.

I think maintenance as is should be discontinued, but the migration should happen gradually, right now it should have become clear that no one is going or can move in one big step.

rtri:
You can try dangling carrots in front of people with features that exist exclusively in develop (and risk doing work that won't be seen by anyone), or you can keep gradually extending maintenance and risk there being nobody left on the engine side by the time this supposed transition is complete.

What is more important to you in the long run?

icexuick:
I've got very limited know-how of the engine and the future of Spring, but all i can say it that we've come so far, we can use all the (frequent) updates (like these) to further build/promote BAR and increase the growing influx of players even more.

I don't know what wise, but i do know that we should not stop developing, and also try to find new paths for the future.. that might be 2 steps forward 1 back.. (..or even more back at first) but i've read so much about develop and 105, i think it's time we really look into this and start to do things...

lhog:
> and risk there being nobody left on the engine side by the time this supposed transition is complete.
> 
> What is more important to you in the long run?

I obviously value your participation in spring, as I think you are aware.

But let me tell one real story on what happened several months when you announced on the mantis or forum that you wouldn't be working on spring actively anymore. The only person who voiced their serious concerns publicly was yours truly. Some refused to believe and suggested to wait, the others were seemingly unaware of what happened and were more concerned that 103/104 had been good enough to move even onto the current maintenance. To be fair several folks were mildly concerned, but to say that world went upside down in one day or one month or half-year would be straight lie. 

Why am I telling this? As weird as it sounds the `nobody left` future doesn't seem to be a valid horror for most of the spring people.

I talked with ZK and BAR leads and I have reasons to believe that no amount of dangling carrots especially as minor as softened particles would justify them to spend man-months on porting their games. First it's a huge effort in itself and secondly there are valid concerns that the OpenGL profile enacted in `develop` is too high, so the visible part of the players will be lost. The facts are such that during your inactive time no one put a single LoC towards moving to develop and I think it's time to accept that as unfortunate fact. What I'm advocating as preparation step is to raise bar gradually based on the `transition` branch based on `maintenance`:
* Release the current maintenance as end-of-the-era release. Conservative games with no manpower can eventually go there if they want to catch up from <=104 times.
* In the `transition` branch, remove FFP/ARB way so the most retarded potatoes finally start saving on daily coffee cup and buy 50$ worth GPU in a week or two.
* Add VAO/VBO primitive so people (BAR and/or ZK realistically) can start porting their code onto the new primitives. Eventually(!) remove or no-op matrix, display lists, gl.Vertex & alike calls.
* In the same time `develop` will likely need to go down to 4.30 level of supported features. Even 4.30 is too high, but I feel like SSBO is too advantageous to skip it.

The above is ofc not nearly as complete and items might come and go as we move. If we move.
Right now it's a stalemate and I have no reasons to believe it will resolve any time soon with the approach you seem to expect to work. Please consider.

P.S. We (I for one) miss you here.

Beherith:
I'm all for moving to develop, unfortunately I probably missed the briefing on what exactly needs to be done, and some general guidelines on how to get there. 
If anyone, Kloot knows how bad my opengl/c code is, but I've worked a lot on lua now, and feel comfortable helping as much as possible to make the transition, since I'm back to devving BAR full time. 
Its always a pleasure to see the great stuff you build into Spring, Kloot, and I wish to make your dream of higher opengl standards come true!

lhog:
screen01795
screen01796

silentwings:
> As weird as it sounds the nobody left future doesn't seem to be a valid horror for most of the spring people.
>I talked with ZK and BAR leads and I have reasons to believe that no amount of dangling carrots especially as minor as softened particles would justify them to spend man-months on porting their games.

I suspect the second part is true and the first part is a misconception caused by the second part. More precisely, I suspect game devs simply do not believe they could successfully port their games.
 
> Release the current maintenance as end-of-the-era release. Conservative games with no manpower can eventually go there if they want to catch up from <=104 times.

+1 from me (since months ago!)

> In the transition branch

I think the only realistic way for the first transition to take place is that (1) "game" side code that was already wholly/mostly written by engine devs (lups, chili, etc) is upgraded _inside of a playable game_ by engine devs, because game devs basically can't do it, and (2) in sync with that, engine devs also help with the major resulting breakages and the process of moving the bigger widgets/gadgets over. If this requires some special engine branch to make the transition through then so be it - holding an opinion on that is beyond my attention level here.

The most obvious approach seems to me that BAR declares a feature freeze on everything except the 105 upgrade, and becomes "the" game on which the transition is based. It is suitable only because it is graphically advanced with experienced devs and few players (afaics, ZK could not freeze, their first call of attention will be their playerbase). Ofc if BAR didn't want it then that's that - if they did want to go for it then the biggest issue seems to me to be that RedUI is crap code and old tech = not worthy of the effort to transition it.

icexuick:
Afaik RedUI is already being replaced by a new FlowUI (from Floris).

I'm probably not the one to make the calls on this end (tech/dev/engine), but as far as i can tell, i think we should go for the '105' » If this can/will solve the problem that we can't get the many ideas/suggestions we have, because most are all 'on hold' due to the 'this is not possible with current-gen-engine'.

lhog:
> +1 from me (since months ago!), why not call it 105.

Is it your powers to make a release and announce it would be a final version of spring as we know it as far as OpenGL is concerned? Sorry for my ignorance I really don't know how responsibilities are split and what's your part in the spring engine development in narrow or wide meaning. So If you can, please do so.

> More precisely, I suspect game devs are either tied to reacting full time to a playerbase and/or simply do not believe they could successfully port their games.

That is more or less what I wrote or meant in the part you didn't quote.

With the rest of your message I think we talk about the same with slightly different angle of view and differences in details. In particular I tend to believe that no game WIP (BAR) or established(ZK) can afford to seize game development and jump over 3 major versions of OpenGL. It would take literally man-months if not years.

And BAR development doesn't work like that: there are at least a dozen contributors doing their pieces, we can't realistically hold them back for months until GL migration is done. Nor we want early players/testers to sit idle without bugfixing. It would be a sure way to lose contributors and potential player base.

I strongly believe the continuous integration is the only way to proceed forward. The development of other parts and testing/gaming by players should not stop for a second. New API from `develop` should appear in the current branch and once the games GL code is moved to the new primitive the gradual switchover should happen.

`develop` itself will certainly need upgrade in some parts (from the top of my head: matrices approach is irritating and inconvient) and downgrade in the other (can't afford to trim off too many players due to high OpenGL requirements).

@GoogleFrog has made a benchmark and the results of `maintenance` vs `develop` performance are encouraging. At least ZK and BAR are poking the ground and want to eventually move to `develop`, however there should be something better than "just rewrite your game in one go".

I suggest whoever considers themselves engine devs to stop idling and act. Not necessarily as per above, but do something. The waiting period is already far longer than it should have been. Alternatively I wouldn't be too surprised if the representatives of the spring game community will setup their own builds of the engine.

silentwings:
> Is it your powers to make a release and announce it would be a final version of spring as we know it as far as OpenGL is concerned?
In terms of pure access rights, yes probably, but since my total engine commits are <100 lines its not something I should do. Hokomoko or abma (or Kloot) are the people to do this.

GoogleFrog:
Zero-K is on https://github.com/spring/spring/commit/89bb8e3d25729638e782a67367cd07ca2f3a7545 (the latest maintenance) with 64 builds on all platforms. It seems to work fine.

abma:
so, can we agree to:

- create 105.0 from maintenance
- drop the maintenance branch
- try to put all efforts in the development branch and focus on creating 106.0?

(i'm not sure what git magic that requires, but it should be possible)

if so, i can make the 105.0 release... for me it atm looks like all active contributers / engine devs are here, so i would see that as a "community decision".

abma:
-> https://springrts.com/phpbb/viewtopic.php?f=12&p=594366

silentwings:
> [from forum] I'm afraid the only right way is to implement things "properly" according to modern best practices. 

Fyi this is what makes me skeptical that a game could transition without it doing a non-gl feature freeze during the transition (or a branch that had one). I guess that whichever new profile is adopted, its not going to come in nicely packaged parts that one can update bit by bit whilst the rest of ogl still works on old stuff, my expectation was that a near-whole update has to be done before the game becomes useable again, with engine being bugfixed alongside. But I guess you/Kloot must have considered this, I'm no ogl expert.

silentwings:
> [from forum] I'm afraid the only right way is to implement things "properly" according to modern best practices. 

Fyi this is what makes me skeptical that a game could transition without it doing a non-gl feature freeze during the transition (or a branch that had one). I guess that whichever new profile is adopted, its not going to come in nicely packaged parts that one can update bit by bit whilst the rest of ogl still works on old stuff, my expectation was that a near-whole update has to be done before the game becomes useable again, with engine being bugfixed alongside. But I guess you/Kloot must have considered this, I'm no ogl expert.

Re "cancerous limbs" I'm not surprised to see projects wanting more control as they grow, and BAR seems on an outwards trajectory to me, but I would bet that the more of the engine you touch, the more you will find _yourself_ wanting independence from games. Abstraction layers don't exist for fun ;)

lhog:
> Fyi this is what makes me skeptical that a game could transition without it doing a non-gl feature freeze during the transition (or a branch that had one).

Not 100% sure either. But we can get closer surely. 



> I guess that whichever new profile is adopted, its not going to come in nicely packaged parts that one can update bit by bit whilst the rest of ogl still works on old stuff,

Idea is that we don't touch a thing how maintenance engine works internally. All necessary internal changes are added to develop already and there is no point to do them once again. Just add a relatively small toolkit on the Lua side, that enables new API primitives. Compatibility profile is called such for reason, as it allows mixing and matching old with new. Then the work can start on the Lua side widget by widget, gadget by gadget. We get away as much as possible from old constructs and replace them with the new. It won't get us performance or anything, just a way to avoid branching out.

> But I guess you/Kloot must have considered this, I'm no ogl expert.
I'm not sure @rtri supports my approach. We used to have a comm channel and talked, but now he's not available. I have rough idea how it should work, but devil is in the details. And I'm sure there 100s of devils on the way forward. Usually it gets in your face only when you start doing stuff.

On the positive note, even if I/we fail, the maintenance engine will not become any worse. We only add stuff to Lua OpenGL, remove nothing, change nothing the way engine works in its core. One might call the proposed additional APIs as "optional".


> Abstraction layers don't exist for fun ;)

I guess the main critique from the game developers, which I happened to voice in the private talk, is that abstraction was too thick. But I don't disagree with you.

rtri:
I was away last week and missed the latest troll-fueled dramaplosion that apparently sprung from this, can't say I'm surprised. However as you will probably understand by now, at this point Spring is mostly a giant source of frustration for me which I have rather little desire to sink myself into any further. Nevertheless, I'll commit (without making explicit promises regarding my time) to the following plan if hokomoko is also onboard:

1) `maintenance` will receive a few more patches (note that this is NOT an invitation for everyone to start promoting their favorite last-minute requests), be released as 105.0, and then terminated
2) `transition` will be created afterwards and receive the GL4 Lua API, *but no other additions of any kind while it exists*
3) *all* new features (such as this PR), fixes, etc will *exclusively* go into and be accepted only for `develop`

Let me know if you are in agreement with these terms or feel like going your own way.

gajop:
Any reason we can't just add the GL4 API to maintenance and treat it as a transition branch that receives fixes? I think our goal should be to make it easy for games to perform the transition.

PS: Anyways, I'm willing to help.

lhog:
@rtri hey man! yeah you missed the whole party.

I fully understand and share your frustration with spring, despite I met very nice support from @abma @silentwings @gajop @ashdnazg and others (thx guys!).

For now I'm contemplating a bit different org approach. In particular we (BAR) would like to take responsibility and lead by example. I will continue the spring branch I started and we plan to use builds of that branch to migrate BAR towards the direction of `develop`. People here have run tests and are impressed by the performance, so something alike to the current `develop` is certainly a goal. 

This decision (org-wise) is done to decrease time we need to communicate and seek for consensus and tighten the feedback loop, so the whole work is limited to one game (not being released, with loyal beta-player base and more importantly a dev team of true collaborators). The results are, however, later on can be re-used by other games, given that spring games share a lot of Lua code.

In the process I plan to change a few things I don't like in `develop`, for example the Lua matrix API, but no downgrade changes are foreseen until we get closer to `transition` - `develop` switchover time.

I cannot promise no additions to `transition` mostly because:

- The moment I started working on the code I realized basic things, like std::string formatting, are missing. Another example could be the use of https://github.com/ThePhD/sol2 for new Lua API modules (for example LuaMatrix::). Things like that are just quality of life and time-to-completion thing and I don't want to suffer for no reason without handy tools. Ofc common modules such as LuaMatrix will go to `develop` with the new libs as well.
- We are at around 4000 deprecated GL LoC so the game migration will certainly take significant time and I don't see why during that some small additions like this one could not be added for the benefit of the game. We are adamant about direction and don't need additional dangling carrots.

Finally I welcome you to join us. We have zero tolerance for politics inside the team and we allow no toxicity. It's all about merit and environments like that makes wonders as I witness: one year of action and we see new talents join and older contributors get active again. The team is excited, productive and people support each other. As miraculous as it sounds after what we all are used to with BA-trolls and stuff, it's true. https://discord.gg/mQ9WDCR

ashdnazg:
I'm on board with Kloot's plan. my time is barely-existent until August, but if we commit to this I will allocate a weekly hour or two to spring to make sure there is no stagnation.

PR #536, opened 2020-06-19
develop-to-maintanance OpenGL API transplantation

lhog:
WIP


gajop:
I recommend we avoid introducing any third party libraries or doing other project changes unrelated to rendering API port in this PR. I'd like to make it simple to merge without asking for similar changes to be first applied to the develop branch.

PR #539, opened 2020-07-16
LuaMatrix

lhog:


rts/Lua/LuaMatrix.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+template<typename TFunc>
+inline void LuaMatrixImpl::AssignOrMultMatImpl(const sol::optional<bool> mult, const bool multDefault, const TFunc& tfunc)
+{
+	const bool multVal = mult.value_or(multDefault);
+	if (!multVal)
+		mat = tfunc();
+	else {
+		CMatrix44f tmpMat = tfunc();
+		mat *= tmpMat;
```
lhog, line 119: Todo: check the order here

rts/Lua/LuaMatrix.cpp
```diff
@@ -0,0 +1,354 @@
...
+inline bool LuaMatrixImpl::ObjectMatImpl(const unsigned int objID, bool mult, sol::this_state L)
+{
+	if (!mult)
+		return GetObjectMatImpl<TObj>(objID, mat, L);
+	else {
+		CMatrix44f mat2;
+		if (!GetObjectMatImpl<TObj>(objID, mat2, L))
+			return false;
+
+		mat *= mat2;
```
lhog, line 146: Todo: check order

rts/Lua/LuaOpenGL.cpp
```diff
@@ -1040,6 +1062,41 @@ inline void LuaOpenGL::CheckDrawingEnabled(lua_State* L, const char* caller)
...
+		if (lua_getstack(L, level, &info)) {
+			lua_getinfo(L, "nSl", &info);
+
+			luaCaller = fmt::format("[{}]:{} Caller: {}", info.short_src, info.currentline, (info.name ? info.name : "<unknown>"));
+		}
+	}
+
+	const auto key = fmt::format("{}{}", caller, luaCaller.c_str());
+
+	if (deprecatedGLWarned.find(key) == deprecatedGLWarned.end()) {
```
ashdnazg, line 1084: early exit

rts/Lua/LuaMatrix.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+#undef near
+#undef far
+
+CMatrix44f LuaMatrixImpl::screenViewMatrix = CMatrix44f{};
+CMatrix44f LuaMatrixImpl::screenProjMatrix = CMatrix44f{};
+
+void LuaMatrixImpl::Zero()
+{
+	std::fill(&mat[0], &mat[0] + 16, 0.0f);
```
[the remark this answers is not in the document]
lhog: I was almost flawless in resisting the temptation of changing something outside of relevant files. I probably can do the change you mentioned, if no one minds.

rts/Lua/LuaMatrix.cpp
```diff
@@ -0,0 +1,351 @@
...
+}
+
+void LuaMatrixImpl::SetMatrixElements(const float m0, const float m1, const float m2, const float m3, const float m4, const float m5, const float m6, const float m7, const float m8, const float m9, const float m10, const float m11, const float m12, const float m13, const float m14, const float m15)
+{
+	const std::array<float, 16> arr { m0, m1, m2, m3, m4, m5, m6, m7, m8, m9, m10, m11, m12, m13, m14, m15 };
```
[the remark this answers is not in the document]
lhog: I think it's MSVC styler that does such thing. If needed I can remove leading and trailing spaces.

rts/Lua/LuaMatrix.hpp
```diff
@@ -0,0 +1,197 @@
...
+struct CSolidObject;
+
+struct LocalModelPiece;
+
+typedef std::tuple< float, float, float, float, float, float, float, float, float, float, float, float, float, float, float, float > tuple16f;
+typedef std::tuple< float, float > tuple2f;
+typedef std::tuple< float, float, float > tuple3f;
+typedef std::tuple< float, float, float, float > tuple4f;
+
+class LuaMatrixImpl {
```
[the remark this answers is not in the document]
lhog: Meant to represent Implementation class. Probably `Impl` can be removed

rts/Lua/LuaMatrix.hpp
```diff
@@ -0,0 +1,197 @@
...
+	{
+		return std::make_tuple(
+			mat[0], mat[1], mat[2], mat[3],
+			mat[4], mat[5], mat[6], mat[7],
+			mat[8], mat[9], mat[10], mat[11],
+			mat[12], mat[13], mat[14], mat[15]
+		);
+	}
+
+	sol::as_table_t<std::vector<float>> GetAsTable()
```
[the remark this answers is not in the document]
lhog: Worth checking

rts/Lua/LuaOpenGL.cpp
lhog: Do you mean: "reverse the condition and quit if true" ?

ashdnazg: yes

rts/Lua/LuaMatrix.cpp
```diff
@@ -0,0 +1,351 @@
...
+		if (!GetObjectMatImpl<TObj>(objID, mat2, L))
+			return false;
+
+		mat *= mat2;
+		return true;
+	}
+}
+
+template<typename TObj>
+inline bool LuaMatrixImpl::GetObjectPieceMatImpl(const unsigned int objID, const unsigned int pieceNum, CMatrix44f& outMat, sol::this_state L)
```
gajop, line 149: is sol::this_state something you're supposed to pass by value?

rts/Lua/LuaMatrix.hpp
```diff
@@ -0,0 +1,197 @@
...
+
+
+struct CSolidObject;
+
+struct LocalModelPiece;
+
+typedef std::tuple< float, float, float, float, float, float, float, float, float, float, float, float, float, float, float, float > tuple16f;
+typedef std::tuple< float, float > tuple2f;
+typedef std::tuple< float, float, float > tuple3f;
+typedef std::tuple< float, float, float, float > tuple4f;
```
gajop, line 37: prefer using over typedef, e.g. : 
`typedef std::tuple< float, float, float, float, float, float, float, float, float, float, float, float, float, float, float, float > tuple16f;`
-> 
`using tuple16f = std::tuple<float, float, float, float, float, float, float, float, float, float, float, float, float, float, float, float>;`

rts/Lua/LuaMatrix.hpp
```diff
@@ -0,0 +1,197 @@
+/* This file is part of the Spring engine (GPL v2 or later), see LICENSE.html */
+
+#ifndef LUA_MATRIX_H
+#define LUA_MATRIX_H
```
gajop, line 4: Filename and ifndef should match, and also our headers should end in .h instead of .hpp for consistency with the rest of the project

rts/Lua/LuaMatrix.hpp
```diff
@@ -0,0 +1,197 @@
...
+				continue;
+
+			f4[i - 1] = maybeValue.value();
+		}
+		f4 = mat * f4;
+		return sol::as_table( std::vector<float> { f4.x, f4.y, f4.z, f4.w } );
+	};
+	LuaMatrixImpl operator* (const LuaMatrixImpl& lmi) const { return LuaMatrixImpl(mat * lmi.GetMatRef()); };
+
+	LuaMatrixImpl operator+ (const LuaMatrixImpl& lmi) const { return LuaMatrixImpl(mat + lmi.GetMatRef()); };
```
gajop, line 152: I don't think this is something we can fix easily, but providing these kind of operators is going to be stupidly inefficient when used from Lua; Every time we apply one of these operations, a new object would have to be created on the heap, which includes temporary variables.

rts/Lua/LuaMatrix.hpp
```diff
@@ -0,0 +1,197 @@
...
+private:
+	template<typename TFunc>
+	inline void AssignOrMultMatImpl(const sol::optional<bool> mult, const bool multDefault, const TFunc& tfunc);
```
gajop, line 155: it makes little sense to declare these functions here seeing as they can't be invoked (templates need to be fully defined to be invoked). Most of these templates can be free functions, either static or in a namespace

rts/Lua/LuaMatrix.hpp
```diff
@@ -0,0 +1,197 @@
...
+	template <typename  TObj>
+	bool ObjectPieceMatImpl(const unsigned int objID, const unsigned int pieceNum, bool mult, sol::this_state L);
+
+	template <typename  TObj>
+	inline static bool IsObjectVisible(const TObj* obj, sol::this_state L);
+
+private:
+	static void SetupScreenMatrices();
+
+	inline static const LocalModelPiece* ParseObjectConstLocalModelPiece(const CSolidObject* obj, const unsigned int pieceNum);
```
gajop, line 176: declaring a function inline without defining it makes little sense (as it cannot be inlined)

rts/Lua/LuaMatrix.hpp
```diff
@@ -0,0 +1,197 @@
...
+	bool ProjectilePieceMatrix(const unsigned int projID, const unsigned int pieceNum, const sol::optional<bool> mult, sol::this_state L) { return ObjectPieceMatImpl<CFeature>(projID, pieceNum, mult.value_or(objectMultDefault), L); } //does this exist at all?
+
+	void ScreenViewMatrix(const sol::optional<bool> mult);
+	void ScreenProjMatrix(const sol::optional<bool> mult);
+
+	void SunViewMatrix(const sol::optional<bool> mult);
+	void SunProjMatrix(const sol::optional<bool> mult);
+
+	void Ortho(const float left, const float right, const float bottom, const float top, const float near, const float far, const sol::optional<bool> mult);
+	void Frustum(const float left, const float right, const float bottom, const float top, const float near, const float far, const sol::optional<bool> mult);
```
gajop, line 111: In function declaration, don't put const on parameters that are passed by value as it has no meaning: pass-by-value is always going to be a copy so the caller cannot distinguish between const and non-const pass-by-value parameters. 

It's perfectly fine to do this in function *definitions* as that describes how a value is used within the function (it mainly helps the reader understand the function body)

rts/Lua/LuaMatrix.hpp
```diff
@@ -0,0 +1,197 @@
...
+#include "Sim/Features/Feature.h"
+#include "Sim/Units/Unit.h"
+
+#undef near
+#undef far
+
+struct lua_State;
+
+
+struct CSolidObject;
```
gajop, line 30: This type of forward declaration be unnecessary as you're already indirectly including it through Sim/Units/Unit.h 

Additionally, as CSolidObject is defined as a class, we should also forward declare it as a class, and not a struct

rts/Lua/LuaMatrix.hpp
```diff
@@ -0,0 +1,197 @@
...
+	inline static bool IsObjectVisible(const TObj* obj, sol::this_state L);
+
+private:
+	static void SetupScreenMatrices();
+
+	inline static const LocalModelPiece* ParseObjectConstLocalModelPiece(const CSolidObject* obj, const unsigned int pieceNum);
+	//static methods
+private:
+	static constexpr bool viewProjMultDefault = false;
+	static constexpr bool objectMultDefault = true;
```
gajop, line 180: capitalize compile constants (CAPITAL_UNDERSCORE notation)

rts/Lua/LuaMatrix.hpp
```diff
@@ -0,0 +1,197 @@
...
+};
+
+class LuaMatrix {
+public:
+	static int GetMatrix(lua_State* L);
+	static bool PushEntries(lua_State* L);
+};
+
+
+#endif //LUA_MATRIX_H
```
gajop, line 197: maybe add a newline here

rts/Lua/LuaMatrix.hpp
```diff
@@ -0,0 +1,197 @@
...
+	//static constants
+private:
+	CMatrix44f mat;
+private:
+	static CMatrix44f screenViewMatrix;
+	static CMatrix44f screenProjMatrix;
+	//private vars
+};
+
+class LuaMatrix {
```
gajop, line 190: Since you're already splitting LuaMatrix into two classes (interface and implementation), it might be a good idea to also fully move LuaMatrixImpl to a different file and only include it in the LuaMatrix.cpp and LuaMatrixImpl.cpp files.
This will reduce the amount of stuff that gets included when we include LuaMatrix.h, and should make compilation faster.

rts/Lua/LuaOpenGL.cpp
```diff
@@ -56,6 +63,7 @@
 #include "Rendering/Textures/NamedTextures.h"
 #include "Rendering/Textures/3DOTextureHandler.h"
 #include "Rendering/Textures/S3OTextureHandler.h"
+
```
gajop, line 66: please avoid adding these newlines

rts/Lua/LuaOpenGL.cpp
```diff
@@ -1040,6 +1062,41 @@ inline void LuaOpenGL::CheckDrawingEnabled(lua_State* L, const char* caller)
...
+		lua_Debug info;
+		const int level = 1; // A calling function from Lua
+		if (lua_getstack(L, level, &info)) {
+			lua_getinfo(L, "nSl", &info);
+
+			luaCaller = fmt::format("[{}]:{} Caller: {}", info.short_src, info.currentline, (info.name ? info.name : "<unknown>"));
+		}
+	}
+
+	const auto key = fmt::format("{}{}", caller, luaCaller.c_str());
```
gajop, line 1082: I assume this key is an std::string? might as well write it out then

rts/Lua/LuaOpenGL.cpp
```diff
@@ -193,7 +214,7 @@ void LuaOpenGL::Free()
 	if (!globalRendering->haveGLSL)
 		return;
 
-	for (const OcclusionQuery* q: occlusionQueries) {
+	for (const auto q: occlusionQueries) {
```
gajop, line 217: why auto here?

rts/Lua/LuaOpenGL.cpp
```diff
@@ -183,6 +196,14 @@ void LuaOpenGL::Init()
...
 
 	canUseShaders = (globalRendering->haveGLSL && configHandler->GetBool("LuaShaders"));
+
+	deprecatedGLWarnLevel = configHandler->GetInt("DeprecatedGLWarnLevel");
+	if (deprecatedGLWarnLevel == 1)
+		deprecatedGLWarned.reserve(64); // only deprecated calls are logged
+	else if (deprecatedGLWarnLevel >= 2)
+		deprecatedGLWarned.reserve(4096); // deprecated calls are logged along with caller information
+
+	deprecatedGLWarned.clear();
```
gajop, line 206: why is clear called here?

rts/Lua/LuaOpenGL.cpp
```diff
@@ -1040,6 +1062,41 @@ inline void LuaOpenGL::CheckDrawingEnabled(lua_State* L, const char* caller)
...
+inline void LuaOpenGL::CondWarnDeprecatedGL(lua_State* L, const char* caller)
+{
+	if (deprecatedGLWarnLevel <= 0)
+		return;
+
+	std::string luaCaller;
+
+	if (deprecatedGLWarnLevel >= 2) {
+		lua_Debug info;
+		const int level = 1; // A calling function from Lua
```
gajop, line 1074: constexpr

rts/Lua/LuaOpenGL.cpp
```diff
@@ -8,10 +8,14 @@
...
-#include "Rendering/GL/myGL.h"
-#include <string>
+//#include "Rendering/GL/myGL.h"
+
 #include <vector>
 #include <algorithm>
+#include <optional>
+
+#include "lib/fmt/format.h"
+#include "lib/sol2/sol.hpp"
```
gajop, line 18: why was sol.hpp included in this file?
also, why was optional included? the only mention of optional I've seen in this PR (outside of this file) are sol::optional and surely that doesn't require the include of the std's header?

rts/Lua/LuaOpenGL.h
```diff
@@ -4,9 +4,14 @@
 #define LUA_GL_H
 
 #include <vector>
+#include <string>
+#include <unordered_set>
 
+#include "Rendering/GL/myGL.h"
 #include "Lua/LuaHandle.h"
 
+#include "Rendering/GL/VBO.h"
```
gajop, line 13: are these Rendering includes necessary to be in the header? whenever possible move them to the .cpp so we reduce compilation time

rts/Lua/LuaMatrix.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+#include "Rendering/Models/3DModel.h"
+#include "Rendering/ShadowHandler.h"
+#include "Rendering/GlobalRendering.h"
+
+#include "Game/Camera.h"
+#include "Game/CameraHandler.h"
+
+#undef near
+#undef far
```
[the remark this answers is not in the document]
ashdnazg: this is necessary due to windows shenanigans

rts/Lua/LuaMatrix.hpp
ashdnazg: another thing I forgot - you don't need GetMatRef here, `lmi.mat` is enough.

rts/Lua/LuaMatrix.cpp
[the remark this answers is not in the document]
sprunk:
```suggestion
#undef far
// necessary due to windows shenanigans
```

rts/Lua/LuaMatrix.cpp
```diff
@@ -0,0 +1,351 @@
...
+			return false;
+
+		mat *= mat2;
+		return true;
+	}
+}
+
+inline void LuaMatrixImpl::SetupScreenMatrices() {
+	// .x := screen width (meters), .y := eye-to-screen (meters)
+	static float2 screenParameters = { 0.36f, 0.60f };
```
sprunk, line 182: 0.36 and 0.60 look somewhat magic

rts/Lua/LuaMatrix.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+CMatrix44f LuaMatrixImpl::screenViewMatrix = CMatrix44f{};
+CMatrix44f LuaMatrixImpl::screenProjMatrix = CMatrix44f{};
+
+void LuaMatrixImpl::Zero()
+{
+	std::fill(&mat[0], &mat[0] + 16, 0.0f);
+}
+
+void LuaMatrixImpl::SetMatrixElements(const float m0, const float m1, const float m2, const float m3, const float m4, const float m5, const float m6, const float m7, const float m8, const float m9, const float m10, const float m11, const float m12, const float m13, const float m14, const float m15)
```
sprunk, line 33: long line is long
```suggestion
void LuaMatrixImpl::SetMatrixElements(
	const float  m0, const float  m1, const float  m2, const float  m3,
	const float  m4, const float  m5, const float  m6, const float  m7,
	const float  m8, const float  m9, const float m10, const float m11,
	const float m12, const float m13, const float m14, const float m15)
```

rts/Lua/LuaMatrix.hpp
```diff
@@ -0,0 +1,197 @@
...
+#include "System/Matrix44f.h"
+#include "System/MathConstants.h"
+#include "System/EventClient.h"
+#include "System/float4.h"
+
+#include "Sim/Features/Feature.h"
+#include "Sim/Units/Unit.h"
+
+#undef near
+#undef far
```
sprunk, line 25: see above
```suggestion
#undef far
// needed due to windows shenanigans
```

rts/Rendering/GlobalRendering.cpp
```diff
@@ -259,6 +269,11 @@ CGlobalRendering::~CGlobalRendering()
 	configHandler->RemoveObserver(this);
 	verticalSync->WrapRemoveObserver();
 
+	delete screenViewMatrix; screenViewMatrix = nullptr;
```
sprunk, line 272: `spring::SafeDelete`? https://github.com/spring/spring/blob/30862626214bd263b1c4489bb197ef1c3dbc0738/rts/System/SafeUtil.h#L20

rts/Rendering/UniformConstants.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+void UniformConstants::Kill()
+{
+	if (!Supported())
+		return;
+
+	for (auto& ucbBuffer : ucbBuffers) {
+		ucbBuffer->Release();
+		delete ucbBuffer;
+		ucbBuffer = nullptr;
```
sprunk, line 46: consider `spring::SafeDelete` (see above)

rts/Rendering/UniformConstants.cpp
```diff
@@ -42,7 +43,7 @@ void UniformConstants::Kill()
 
 	for (auto& ucbBuffer : ucbBuffers) {
 		ucbBuffer->Release();
-		delete ucbBuffer;
+		spring::SafeDelete(ucbBuffer);
 		ucbBuffer = nullptr;
```
sprunk, line 47: no longer needed
```suggestion
```

rts/Lua/LuaMatrixImpl.cpp
```diff
@@ -0,0 +1,323 @@
...
+}
+
+void LuaMatrixImpl::SetMatrixElements(
+	const float  m0, const float  m1, const float  m2, const float  m3,
+	const float  m4, const float  m5, const float  m6, const float  m7,
+	const float  m8, const float  m9, const float m10, const float m11,
+	const float m12, const float m13, const float m14, const float m15)
+{
+	const std::array<float, 16> arr {m0, m1, m2, m3, m4, m5, m6, m7, m8, m9, m10, m11, m12, m13, m14, m15};
+	std::copy(arr.cbegin(), arr.cend(), &mat.m[0]);
```
gajop, line 36: I've a feeling CMatrix44f could use a constructor for this.
Something akin to 
```
    CMatrix44f(const float  m0, const float  m1, const float  m2, const float  m3,
	const float  m4, const float  m5, const float  m6, const float  m7,
	const float  m8, const float  m9, const float m10, const float m11,
	const float m12, const float m13, const float m14, const float m15) : m{m0, m1, m2, m3, m4, m5, m6, m7, m8, m9, m10, m11, m12, m13, m14, m15}
    {}
```
Then you can just
```
mat = CMatrix44f { m0, m1, m2, m3, m4, m5, m6, m7, m8, m9, m10, m11, m12, m13, m14, m15 };
```

rts/Lua/LuaMatrixImpl.cpp
```diff
@@ -0,0 +1,323 @@
...
+}
+
+void LuaMatrixImpl::Scale(const float s)
+{
+	mat.Scale({s, s, s});
+}
+
+void LuaMatrixImpl::RotateRad(const float rad, const float x, const float y, const float z)
+{
+	// TODO: Find out why minus and if it is correct.
```
gajop, line 56: Please do find out before we merge :)

rts/Lua/LuaMatrixImpl.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+template<typename TObj>
+bool LuaMatrixImpl::IsObjectVisible(const TObj* obj, sol::this_state L)
+{
+	if (CLuaHandle::GetHandleFullRead(L))
+		return true;
+
+	const int readAllyTeam = CLuaHandle::GetHandleReadAllyTeam(L);
+
+	if ((readAllyTeam < 0) && (readAllyTeam == CEventClient::NoAccessTeam))
```
gajop, line 98: parentheses aren't necessary

rts/Lua/LuaMatrixImpl.cpp
```diff
@@ -0,0 +1,323 @@
...
+	if (obj == nullptr)
+		return nullptr;
+
+	if (!IsObjectVisible(obj, L))
+		return nullptr;
+
+	return obj;
+}
+
+const LocalModelPiece* LuaMatrixImpl::ParseObjectConstLocalModelPiece(const CSolidObject* obj, const unsigned int pieceNum)
```
gajop, line 130: this method can be const I think

rts/Lua/LuaMatrixImpl.cpp
```diff
@@ -0,0 +1,323 @@
...
+		return false;
+
+	if constexpr(std::is_same<TObj, CProjectile>::value) //CProjectile has no IsInLosForAllyTeam()
+		return losHandler->InLos(obj, readAllyTeam);
+	else //else is needed. Otherwise won't compile
+		return obj->IsInLosForAllyTeam(readAllyTeam);
+}
+
+template<typename TObj>
+const TObj* LuaMatrixImpl::ParseObject(int objID, sol::this_state L)
```
gajop, line 108: maybe this can be const too

rts/Lua/LuaMatrixImpl.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+bool LuaMatrixImpl::ProjectileMatrix(const unsigned int projID, const sol::optional<bool> mult, sol::this_state L)
+{
+	return ObjectMatImpl<CProjectile>(projID, mult.value_or(OBJECT_MULT_DEFAULT), L);
+}
+
+///////////////////////////////////////////////////////////
+
+template<typename TObj>
+bool LuaMatrixImpl::IsObjectVisible(const TObj* obj, sol::this_state L)
```
gajop, line 91: const as well?

rts/Lua/LuaMatrixImpl.cpp
```diff
@@ -0,0 +1,323 @@
...
+void LuaMatrixImpl::AssignOrMultMatImpl(const sol::optional<bool> mult, const bool multDefault, const CMatrix44f* matIn)
+{
+	const bool multVal = mult.value_or(multDefault);
+
+	if (!multVal) {
+		mat = *matIn;
+		return;
+	}
+
+	//TODO: check multiplication order
```
gajop, line 158: check before merge

rts/Lua/LuaMatrixImpl.cpp
```diff
@@ -0,0 +1,323 @@
...
+}
+
+void LuaMatrixImpl::Frustum(const float left, const float right, const float bottom, const float top, const float _near, const float _far, const sol::optional<bool> mult)
+{
+	AssignOrMultMatImpl(mult, LuaMatrixImpl::VIEWPROJ_MULT_DEFAULT, CMatrix44f::ClipPerspProj(left, right, bottom, top, _near, _far, globalRendering->supportClipSpaceControl * 1.0f));
+}
+
+void LuaMatrixImpl::CameraViewMatrix(const sol::optional<unsigned int> cameraIdOpt, const sol::optional<bool> mult)
+{
+	const auto cam = GetCamera(cameraIdOpt);
```
gajop, line 271: here and in other methods:
`auto cam` -> `CCamera* cam`
`auto matIn` -> `CMatrix44f& matIn`

rts/Lua/LuaMatrixImpl.h
```diff
@@ -0,0 +1,164 @@
...
+
+	void CameraViewMatrix(const sol::optional<unsigned int> cameraIdOpt, const sol::optional<bool> mult);
+	void CameraProjMatrix(const sol::optional<unsigned int> cameraIdOpt, const sol::optional<bool> mult);
+	void CameraViewProjMatrix(const sol::optional<unsigned int> cameraIdOpt, const sol::optional<bool> mult);
+	void CameraViewInverseMatrix(const sol::optional<unsigned int> cameraIdOpt, const sol::optional<bool> mult);
+	void CameraProjInverseMatrix(const sol::optional<unsigned int> cameraIdOpt, const sol::optional<bool> mult);
+	void CameraViewProjInverseMatrix(const sol::optional<unsigned int> cameraIdOpt, const sol::optional<bool> mult);
+	void CameraBillboardMatrix(const sol::optional<unsigned int> cameraIdOpt, const sol::optional<bool> mult);
+
+	tuple16f GetAsScalar()
```
gajop, line 89: const; last time I repeat this; please add it to other methods

gajop, line 89: btw, this and some other methods can also have a [[nodiscard]] attribute as it makes no sense to invoke them without using the return value

rts/Lua/LuaMatrixImpl.h
```diff
@@ -0,0 +1,164 @@
...
+		});
+	}
+
+public:
+	const CMatrix44f& GetMatRef() const { return  mat; }
+	const CMatrix44f* GetMatPtr() const { return &mat; }
+public:
+	sol::as_table_t<std::array<float, 4>> operator* (const sol::table& tbl) const {
+		float4 f4 {0.0f, 0.0f, 0.0f, 1.0f};
+		for (auto i = 1; i <= 4; ++i) {
```
gajop, line 115: please avoid auto for primitive types

rts/Lua/LuaMatrixImpl.h
```diff
@@ -0,0 +1,164 @@
...
+			const sol::optional<float> maybeValue = tbl[i];
+			if (!maybeValue)
+				continue;
+
+			f4[i - 1] = maybeValue.value();
```
gajop, line 120: might as well write this as 
`f4[i - 1] = tbl[i].value_or(f4[i - 1]);`
saves you an if-check and a few lines (am I nitpicking?)

rts/Lua/LuaMatrixImpl.h
```diff
@@ -0,0 +1,164 @@
...
+		}
+		f4 = mat * f4;
+		return sol::as_table(std::array<float, 4> {f4.x, f4.y, f4.z, f4.w});
```
gajop, line 123: why can't we use our types: float4, float3, CMatrix44f instead of std::array? seems like a lot of unnecessary copy-pasting going around

rts/Rendering/GlobalRendering.cpp
```diff
@@ -250,6 +258,9 @@ CGlobalRendering::CGlobalRendering()
 	, sdlWindows{nullptr, nullptr}
 	, glContexts{nullptr, nullptr}
 {
+	screenViewMatrix = new CMatrix44f{};
+	screenProjMatrix = new CMatrix44f{};
```
gajop, line 262: we really should use unique_ptr :P

gajop: (it's not something I expect you to address in this PR; the codebase just doesn't use it for some reason)

rts/Rendering/UniformConstants.cpp
```diff
@@ -0,0 +1,132 @@
...
+	{
+		GLint uniformBufferOffset = 0;
+
+		glGetIntegerv(GL_UNIFORM_BUFFER_OFFSET_ALIGNMENT, &uniformBufferOffset);
+
+		umbBufferSize = ((sizeof(UniformMatricesBuffer) / uniformBufferOffset) + 1) * uniformBufferOffset;
+		upbBufferSize = ((sizeof(UniformParamsBuffer) / uniformBufferOffset) + 1) * uniformBufferOffset;
+	}
+
+	for (auto& buf : umbBuffers) {
```
gajop, line 33: as usual, avoid auto

rts/Lua/LuaMatrixImpl.cpp
lhog: error: static member function 'static const LocalModelPiece* LuaMatrixImpl::ParseObjectConstLocalModelPiece(const CSolidObject*, unsigned int)' cannot have cv-qualifier

rts/Lua/LuaMatrixImpl.cpp
lhog: it's a static method too

rts/Lua/LuaMatrixImpl.cpp
lhog: Same as above. it's static.

rts/Lua/LuaMatrixImpl.h
lhog: These methods are effectively only used from Lua.
I'm not sure if `[[nodiscard]]` makes any sense here, because it enforces nothing on the Lua side.

rts/Rendering/UniformConstants.cpp
lhog: This led to the following fugliness: `for (VBO*& buf : umbBuffers) {`

rts/Lua/LuaMatrixImpl.h
lhog: I implemented the requested change, but unfortunately in this case values are stored as usertypes.
I'm gonna try to get some advise how to deal with it, but for now I'm reverting this change

gajop: Are you sure about that? I thought this would trigger the container&& overload and basically copy the content to a table. My guess is we'd need to satisfy some sort of API (e.g. std::size and overload operator[] or provide a std::begin and std::end implementations)

lhog: > Are you sure about that?

Checked this from Lua, The use of `float4` and `CMatrix44f` caused the creation of the userdata object.

> My guess is we'd need to satisfy some sort of API (e.g. std::size and overload operator[] or provide a std::begin and std::end implementations)

Looks like it's indeed the case. I'm sure at least about `.size()` requirement as I tried to also expose CMatrix44f.m and got `.size()` missing as one of the compilation errors.
Any idea what to do?

gajop: changing `sol::as_table_t<std::array<float, 4>>` to `sol::as_table_t<float4>` in a few places (this function & in LuaMatris.cpp's sol:meta_function part (line 80)) is enough to make it compile: https://godbolt.org/z/9ahEqf . spring crashes on start due to that other issue you mentioned
The same can be done for the GetAsTable, like so:
```
	sol::as_table_t<CMatrix44f> GetAsTable()
	{
		return sol::as_table(mat);
	}
```

but I'm surprised this works as CMatrix44f doesn't seem to satisfy their requirement for a container (https://sol2.readthedocs.io/en/latest/api/as_table.html): "The container passed here can be a pointer, a reference, a std::reference_wrapper around a container, or just a plain container value. It must have a begin/end function, and if it has a std::pair<Key, Value> as its value_type, it will be pushed as a dictionary. Otherwise, it’s pushed as a sequence."

It doesn't seem to have an .begin() and .end() function. If it turns out that things are just compiling, but not working, I recommend you add a .begin() and .end() functions that would point to the first and after-last member of the matrix and vector

rts/Rendering/UniformConstants.h
```diff
@@ -0,0 +1,80 @@
...
+	static UniformConstants& GetInstance() {
+		static UniformConstants uniformConstantsInstance;
+		return uniformConstantsInstance;
+	};
+	static bool Supported() {
+		static bool supported = GLEW_ARB_uniform_buffer_object && GLEW_ARB_shading_language_420pack; //UBO && UBO layout(binding=x)
+		return supported;
+	}
+public:
+	UniformConstants() : umbBufferMap(nullptr), upbBufferMap(nullptr), umbVBO(nullptr), upbVBO(nullptr), umbBufferSize(0), upbBufferSize(0) {};
```
gajop, line 52: `UniformConstants() : umbBufferMap(nullptr), upbBufferMap(nullptr), umbVBO(nullptr), upbVBO(nullptr), umbBufferSize(0), upbBufferSize(0) {};`
->
1) prefer to just initialize these values in the class definition as they don't depend on the constructor parameters
2) further, you don't want to initialize unique_ptr explicitly, it will be nullptr by default
3) and lastly, take care of your initialization order; you should be writing member initialization in the order you've defined them in the class definition, because C++ will actually initialize them in that order (and not the order you've written it), so you might cause a non-obvious bug if there's a dependency between your members

PR #543, opened 2020-09-16
Draw build call in

Helwor:
Getter/Setter CallIn controlling the main map drawing of build grid and build ghost (don't change build grid drawn on minimap).
For the purpose of removing them when one want to draw customized elevation of ghost/grid along with auto terraforming widget.
usage:
Spring.SetDrawBuild(bool grid, bool ghost)
Spring.GetDrawBuild() -> bool grid, bool ghost

### DIFF
--- a/rts/Game/UI/GuiHandler.cpp
+++ b/rts/Game/UI/GuiHandler.cpp
@@ -3936,7 +3936,8 @@ void CGuiHandler::DrawMapStuff(bool onMiniMap)
 		assert(buildeeDef != nullptr);
 
 		if (!buildInfoQueue.empty()) {
-			unitDrawer->SetupShowUnitBuildSquares(onMiniMap, true);
+			if (drawBuildGrid || onMiniMap)
+				unitDrawer->SetupShowUnitBuildSquares(onMiniMap, true);
 
 			// first pass; grid squares
 			for (const BuildInfo& bi: buildInfoQueue) {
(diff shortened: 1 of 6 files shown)


rts/Game/UI/GuiHandler.cpp
```diff
@@ -3936,7 +3936,8 @@ void CGuiHandler::DrawMapStuff(bool onMiniMap)
 		assert(buildeeDef != nullptr);
 
 		if (!buildInfoQueue.empty()) {
-			unitDrawer->SetupShowUnitBuildSquares(onMiniMap, true);
+			if (drawBuildGrid || onMiniMap)
```
gajop, line 3939: again, what's the purpose of onMiniMap here?

rts/Game/UI/GuiHandler.cpp
```diff
@@ -3936,7 +3936,8 @@ void CGuiHandler::DrawMapStuff(bool onMiniMap)
...
+				unitDrawer->SetupShowUnitBuildSquares(onMiniMap, true);
```
gajop, line 3940: this isn't your fault, but SetupShowUnitBuildSquares is really an awful function: the second argument is always known at compile time so we can easily avoid unnecessary calls entirely, and the first parameter is not actually used in the function..

rts/Game/UI/GuiHandler.cpp
```diff
@@ -3954,26 +3955,28 @@ void CGuiHandler::DrawMapStuff(bool onMiniMap)
 						}
 					}
 				}
-
-				if (unitDrawer->ShowUnitBuildSquares(bi, buildCommands, true)) {
-					buildColors.emplace_back(0.7f, 1.0f, 1.0f, 0.4f); // yellow
-				} else {
-					buildColors.emplace_back(1.0f, 0.5f, 0.5f, 0.4f); // red
+				if (drawBuildGrid || onMiniMap){
```
gajop, line 3958: same comments about style here

gajop (changes requested):

rts/Lua/LuaUnsyncedCtrl.cpp
```diff
@@ -2769,6 +2771,15 @@ int LuaUnsyncedCtrl::SetDrawSelectionInfo(lua_State* L)
 /******************************************************************************/
 /******************************************************************************/
 
+int LuaUnsyncedCtrl::SetDrawBuild(lua_State* L)
+{
+	if (guihandler != nullptr){
+			guihandler->drawBuildGrid = !!luaL_checkboolean(L, 1);
+			guihandler->drawBuildGhost = !!luaL_checkboolean(L, 2);
+	}
+	return 0;
```
sprunk, line 2780: Why is this different than the `maintenance` version? I don't see any difference but this will make future cherry-picks harder.
```suggestion
	if (guihandler != nullptr)
		return 0;

	guihandler->drawBuildGrid = luaL_checkboolean(L, 1);
	guihandler->drawBuildGhost = luaL_checkboolean(L, 2);
	return 0;
```
 (Also, no need for the `!!` since `luaL_checkboolean` is already a bool, but this is already not present in the `maintenance` version.)

rts/Game/UI/GuiHandler.cpp
Helwor: I lack the knowledge yet to better that function; however about adding 'onMinimap' it's the same as in maintenance version. It is needed to distinguish minimap grid drawing and main window drawing.

rts/Game/UI/GuiHandler.cpp
Helwor: Same as in maintenance version. It is needed to distinguish minimap grid drawing and main window drawing.

rts/Game/UI/GuiHandler.cpp
Helwor: fixing it

rts/Lua/LuaUnsyncedCtrl.cpp
Helwor: fixing that

abma:
my opengl knowledge is way to limited, but my impression is, that the minimap should be at least partly implemented as lua and not more config settings / code paths added to the drawing code.

sorry, i'm just trying to give feedback with my limited knowledge: having this open for 16 months was bad :-|

PR #545, opened 2020-10-12
add github webhook

TurBoss:
setting this in  github settings 

Payload url:
http://buildbot.springrts.com/change_hook/github?project=spring/spring

Content type:
x-www-form-urlencoded
and create a secret phrase


abma:
great, thanks!

PR #548, opened 2020-11-25
Fix building errors for non-prefixed docbook versions

sanguinariojoe:
In my manjaro docbook-xsl version is 1.79.2, which is not considered [in the prefixed list](https://github.com/spring/spring/blob/develop/rts/build/cmake/FindAsciiDoc.cmake#L38), so it fallbacks to [file based search](https://github.com/spring/spring/blob/develop/rts/build/cmake/FindAsciiDoc.cmake#L51).

Unfortunately, if I manually type the list command I get 2 files:

```
$ ls /usr/share/xml/docbook/xsl-stylesheets-*/manpages/docbook.xsl
/usr/share/xml/docbook/xsl-stylesheets-1.79.2/manpages/docbook.xsl  /usr/share/xml/docbook/xsl-stylesheets-1.79.2-nons/manpages/docbook.xsl
```

Driving to errors. This PR fix that error, getting just the first found file. Tested just on maintenance version in my current computer.


abma:
thank you!

PR #549, opened 2020-11-30
fix invalid unsigned arithmetic.

ihsinme:
your check is incorrect. since the variables are unsigned, it is equivalent to len != ofs.
so I suggest a simple fix for the error.


sprunk:
It's not incorrect because `ofs <= len` always.

PR #550, opened 2021-01-01
Update cleanup.sh so that testversions of engine kept longer for download

saturnvgoesup:
do not delete so soon.
relevant forum thread: https://springrts.com/phpbb/viewtopic.php?f=1&t=40040

### DIFF
--- a/buildbot/master/cleanup.sh
+++ b/buildbot/master/cleanup.sh
@@ -3,7 +3,7 @@
 # and debug symbols older than ${DBG_AGE}.
 # ${DBG_AGE} must be smaller than or equal to ${AGE}.
 
-AGE=120
+AGE=9999
 DBG_AGE=30
 ZIP_AGE=10
 


abma:
the disk space on springrts.com is limited, there must be some autodeletion.

something similar like

https://github.com/spring/pr-downloader/blob/master/scripts/cleanup-builds.py 

needs to be used which deletes by "short" age and keeps the last n files.

saturnvgoesup:
there still is autodeletion, just slower.

used disk space depends on the ratio between new build and time until deletion.
in past this ratio was such, that all builds were deleted. no build at all remained for download.
i suggest to go with the dirty solution of longer deletion-time and adjust it if disk space should become a problem.

abma:
>  i suggest to go with the dirty solution of longer deletion-time and adjust it if disk space should become a problem.

no, as when springrts.com runs out of disk space the database very likely will corrupt. this is much worse than not having development builds.

saturnvgoesup:
the forum thread got locked right now and so i lost interest. you can close this, too. in fact close down everything.

abma:
this pull request comes way to late, sorry. when "fixing" it, please fix it orderly and not quick and dirty as it will create much more worse problems.

will you care about enough free disk space on springrts.com? i think no: so you have to listen to me.

saturnvgoesup:
we had both edited our posts at same time with the idea of keeping a constant number of builds. that would have keep disk space usage constant.
the latest forum posts reminded me that the issues are deeper and not related to code.

abma:
a constant number of builds would lead to the problem, that in a commit spree builds would be deleted to soon:

ideally the script deletes based on disk space and keeps at least n builds and auto deletes very old builds.

it comes to the point that the script as it is, isn't this bad: we just need to make more often a stable release.

saturnvgoesup:
as I said: "the latest forum posts reminded me that the issues are deeper and not related to code."

you can fantasize about the ideal delete-script but you lost focus on reality.
even re-uploading the latest builds by hand would have been a solution to prevent all the dead hosts. people had posted their backups so there was not even need to compile again.

why were games and players even pushed towards using test-builds? why the ban of stable 103.0 when the ban of 98.0 had already shown what happens: you can not force players to use certain engine versions. they might instead just leave. 
that is no way to run a multiplayer game.

PR #551, opened 2021-01-11
Set SUBSYSTEM CONSOLE for Windows system

Yaribz:
This allows Windows users to actually see the output of spring-dedicated.exe without having to redirect it.
This is already done correctly for spring-headless, but for some reason it seems it was forgotten for spring-dedicated.


gajop:
Seems good, I'll merge .. to develop.

As `maintenance` is deleted and `develop` is hardly useful for anything, what are we to do with this?
Do we use the `transition` branch (pretend real hard it's not just a rename for `maintenance`), and continue providing a *working* version of spring there?

PR #552, opened 2021-03-04
Add `isSubmarine` to unitdefs' movedef subtable

sprunk:


lhog:
Is this one needed/wanted still?

sprunk:
Yes but it should probably be PR'd to `transition` instead, and/or to BAR.

lhog:
I'm unsure either. I guess the final decision is on Kloot as the one who put tremendous effort in spring in general and on develop specifically.

rtri:
FTR I don't anticipate much non-rendering work being done for the foreseeable future, so for PR's following the usual develop->transition path is fine for now.

PR #553, opened 2021-04-13
Unit/Feature defs loading: accept `health`

sprunk:
More natural and it's also the key under which the value later gets exposed to wupgets.
`maxDamage` (or `damage` for features) still accepted because it's the TA spelling and what is currently used.


gajop (approved):

gajop:
thanks!

PR #558, opened 2021-04-22
Provide PCG32::result_type

AMDmi3:
Provide result_type member type in PCG32, which makes it compatible with C++ UniformRandomBitGenerator and, as result, makes it usable in std::shuffle (which deprecated std::random_shuffle calls in Spring should be replaced with).

### DIFF
--- a/rts/System/GlobalRNG.h
+++ b/rts/System/GlobalRNG.h
@@ -46,6 +46,8 @@ public:
 	typedef uint32_t res_type;
 	typedef uint64_t val_type;
 
+	using result_type = res_type;
+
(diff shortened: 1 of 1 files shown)


sprunk:
```
/home/travis/build/spring/spring/rts/System/GlobalRNG.h:49:22: error: unknown type name 'rng_res_type'; did you mean 'res_type'?

        using result_type = rng_res_type;

                            ^~~~~~~~~~~~

                            res_type

/home/travis/build/spring/spring/rts/System/GlobalRNG.h:46:19: note: 'res_type' declared here

        typedef uint32_t res_type;

                         ^
```

AMDmi3:
Fixed. This was originally a patch against 105.0, hence the difference.

abma:
thanks you!

PR #559, opened 2021-05-12
This is to merge back BAR branch into spring/transition

lhog:
This revision of the branch should have near 100% compatibility with 105.


rtri:
There are too many changes in too many places here for me (or anyone) to merge as-is. I'll review this first if you don't mind.

lhog:
Oh hey Kloot!

Sure, I warned that the amount would soon be unmanageable several months ago and indeed it has become such. Sadly back in the day I hadn't seen any interest to get it merged. Hopefully these days it's gonna be different.

I want to continue what started as BAR branch at spring/spring as per https://springrts.com/phpbb/viewtopic.php?p=598472#p598472

rtri:
Fine with me, so long as you can accept and respect other people's reduced availability since (at least on my end) working from home has sadly not created more free time to spend elsewhere. Tomorrow is a national holiday around these parts so I'll dive into this then.

lhog:
It's fine, take your time. The life has not become easier for many of us since the pandemic started.
I personally struggle to share the time between the work, family, spring/BAR and recreation.

I also want to set certain expectations from my side:
* I'm not a programmer by trades, nor I have dealt with C++ largely before. I learn as I go, so any coding advises from you, @gajop , @ashdnazg or anyone proficient are welcome.
* With PR of such size, I hope to not see remarks like number of spaces, new lines, etc. Some glaring style violations (if any) can be fixed of course.
* Since I was asked to keep BAR way of migration up to date, I'll need commit rights to spring/spring.

rts/Map/ReadMap.h
```diff
@@ -276,6 +281,8 @@ class CReadMap
 	float2 initHeightBounds; //< initial minimum- and maximum-height (before any deformations)
 	float2 currHeightBounds; //< current minimum- and maximum-height
 
+	std::map<const float, int> heightRefMap;
```
rtri, line 284: There are no comments here, so I don't understand why you want to keep a count of every unique height value at all. What was wrong with tracking `currHeightBounds` the old way and why are refcounts the only possible fix? `UpdateHeightsRefMap` adds a ton of indirection (hidden via `std::map`, which is a terrible data structure in practice if you care about performance) and branching to a rather critical code path; this is **BAD** because the heightmap will be getting absolutely bombarded with updates during large-scale late-game play. Please reconsider your solution.

rts/Map/ReadMap.cpp
```diff
@@ -258,6 +263,19 @@ void CReadMap::PostLoad()
 		mipPointerHeightMaps[i] = &mipCenterHeightMaps[i - 1][0];
 	}
 	*/
+
+	mipPointerHeightMaps.fill(nullptr);
```
rtri, line 267: The commented-out code right above can now be removed.

rts/Rendering/Env/CubeMapHandler.cpp
```diff
@@ -62,8 +62,8 @@ bool CubeMapHandler::Init() {
 		glBindTexture(GL_TEXTURE_CUBE_MAP, envReflectionTexID);
 		glTexParameteri(GL_TEXTURE_CUBE_MAP, GL_TEXTURE_MIN_FILTER, generateMipMaps? GL_LINEAR_MIPMAP_LINEAR: GL_LINEAR);
 		glTexParameteri(GL_TEXTURE_CUBE_MAP, GL_TEXTURE_MAG_FILTER,                                           GL_LINEAR); // magnification doesn't use mips
-		glTexParameteri(GL_TEXTURE_CUBE_MAP, GL_TEXTURE_WRAP_S, globalRendering->supportSeamlessCubeMaps? GL_MIRRORED_REPEAT: GL_CLAMP_TO_EDGE);
```
rtri, line 65: Why?

rts/Rendering/GL/VAO.h
```diff
@@ -0,0 +1,48 @@
...
+#define GL_VAO_H
+
+#include <cstdint>
+#include <utility>
+#include "Rendering/GL/myGL.h"
+
+struct VAO {
+public:
+	static bool IsSupported() {
+		static bool supported = GLEW_ARB_vertex_array_object;
```
rtri, line 13: Why `static bool`? The cost of looking up an exported GLEW symbol is marginal.

rts/Rendering/GL/VAO.cpp
```diff
@@ -0,0 +1,3 @@
+/* This file is part of the Spring engine (GPL v2 or later), see LICENSE.html */
+
+#include "VAO.h"
```
rtri, line 3: Why is this file empty? GL calls should not live in headers too much.

rtri (commented):
Finished an initial "quick" review pass, 184 / 230 files viewed so far. The others (the meat of this PR and some of the diffs I prechecked here) will need much more time to chew / digest I'm afraid.

rts/Game/Game.cpp
```diff
@@ -1322,14 +1338,17 @@ bool CGame::Draw() {
 		// (unlikely); the minimap update also depends on GenerateIBLTextures for unbinding its FBO
 		worldDrawer.GenerateIBLTextures();
 
-		camera->Update();
+		//camera->Update();
```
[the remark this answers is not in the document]
lhog: Yes, camera is supposed to updated once per draw frame. So as uniform buffer holding necessary matrices.

rts/System/Main.cpp
```diff
@@ -17,11 +17,16 @@
...
 #include <clocale>
 #include <cstdlib>
+#include <cstdint>
 
 #ifdef _WIN32
 	#include "lib/SOP/SOP.hpp" // NvOptimus
 #endif
 
+// https://stackoverflow.com/a/27881472/9819318
+EXTERNALIZER_B EXPORT_CLAUSE uint32_t NvOptimusEnablement = 0x00000001;         EXTERNALIZER_E //Optimus/NV use discrete GPU hint
```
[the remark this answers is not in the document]
lhog: Probably not many. It was an attempt to force drivers to use discrete GPU upon one of the users request. The request was fulfilled in hope that some drivers will take a note. It's not hurting performance or anything (though likely doesn't do much).

rts/Sim/Units/CommandAI/MobileCAI.cpp
```diff
@@ -615,9 +615,10 @@ void CMobileCAI::ExecuteGuard(Command& c)
 		return;
 	}
 
+	constexpr int retaliationTimeout = 40;
 	const bool pushAttackCommand =
 		owner->unitDef->canAttack &&
-		(guardee->lastAttackFrame + 40 < gs->frameNum) &&
+		(guardee->lastAttackFrame + retaliationTimeout > gs->frameNum) &&
```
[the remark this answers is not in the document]
lhog: Someone reported it was broken on Discord, showed the demo and I was able to quick-fix it.

P.S. Can we bury ancient Mantis in favor of github issues?

.gitmodules
```diff
@@ -18,7 +9,11 @@
 	url = git://github.com/spring/pyunitsync.git
 [submodule "tools/pr-downloader"]
 	path = tools/pr-downloader
-	url = git://github.com/spring/pr-downloader.git
+	url = https://github.com/beyond-all-reason/pr-downloader.git
```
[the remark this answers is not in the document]
lhog: There were some removals going on spring/pr-d. I wasn't sure why and replaced it just in case.
We haven't developed anything for pr-d. I'll replace it back.

rts/System/FileSystem/ArchiveScanner.cpp
```diff
@@ -1307,7 +1307,7 @@ std::vector<std::string> CArchiveScanner::GetAllArchivesUsedBy(const std::string
 
 			if (aii == archiveInfosIndex.end()) {
 				if (!HandleUnresolvedDep(lwrCaseName))
-					throw content_error("Dependent archive \"" + lwrCaseName + "\" not found");
+					throw content_error("Dependent archive \"" + resolvedName + "\" not found");
```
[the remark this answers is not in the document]
lhog: `resolvedName` is communicated mostly for someone copy-pasting it from infolog to pr-d.
I don't think `lwrCaseName` has any significant meaning outside of being internal key. No?

rts/Sim/Misc/SmoothHeightMesh.cpp
```diff
@@ -405,8 +359,35 @@ void SmoothHeightMesh::MakeSmoothMesh()
...
+			sum += 2.0f * gaussianKernel[i];
+		}
+
+		for (auto& gk : gaussianKernel) {
+			gk /= sum;
+		}
+	};
+
+	constexpr float gSigma = 5.0f;
+	std::vector<float> gaussianKernel;
```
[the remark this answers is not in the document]
lhog: AirMesh should get updated upon heightmap change. This is in TODO, but idk when I can attend to it.
For now previously broken mesh is fixed, but it won't stay relevant if the map got deformed or terraformed. 

With your permission I leave it as is, and redo when partial updates will be implemented.

rts/Rendering/Env/GrassDrawer.cpp
```diff
@@ -222,6 +210,19 @@ CGrassDrawer::CGrassDrawer()
 		grassMap.resize(mapDims.mapx * mapDims.mapy / (grassSquareSize * grassSquareSize));
 		memcpy(grassMap.data(), grassdata, grassMap.size());
 		readMap->FreeInfoMap("grass", grassdata);
+
```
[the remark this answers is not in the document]
lhog: @Beherith Can you please do that?

rts/Rendering/GlobalRendering.cpp
```diff
@@ -690,34 +706,46 @@ void CGlobalRendering::SetGLSupportFlags()
...
-		// but those can't be rendered to)
-		GLint state = 0;
-		glTexImage2D(GL_PROXY_TEXTURE_2D, 0, GL_DEPTH_COMPONENT24, 16, 16, 0, GL_LUMINANCE, GL_FLOAT, nullptr);
-		glGetTexLevelParameteriv(GL_PROXY_TEXTURE_2D, 0, GL_TEXTURE_WIDTH, &state);
-		support24bitDepthBuffer = (state > 0);
+	//stick to the theory that reported = exist
+	//supportMSAAFrameBuffer &= ((globalRenderingInfo.glContextVersion.x * 10 + globalRenderingInfo.glContextVersion.y) >= 32);
+
+	int iter = 0;
+	for (const int bits : {0, 16, 24, 32}) {
```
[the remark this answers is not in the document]
lhog: Probably. 0 means testing against GL_DEPTH_COMPONENT.
Since we are aiming at GL4.2/4.3 it's probably not needed.

TODO.

rts/Lua/LuaUnsyncedCtrl.cpp
```diff
@@ -851,13 +851,21 @@ int LuaUnsyncedCtrl::SetCameraState(lua_State* L)
...
 	camHandler->CameraTransition(luaL_optfloat(L, 2, 0.0f));
 
-	const bool retval = camHandler->SetState(ParseCamStateMap(L, 1));
+	CCameraController::StateMap camState;
+	if (hasState)
+		camState = ParseCamStateMap(L, 1);
+	else
+		camHandler->GetState(camState);
+
+	const bool retval = camHandler->SetState(camState);
```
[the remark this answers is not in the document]
lhog: IMO it hurts readability. 
Also I was going to change SetCameraState() implementation because of https://github.com/beyond-all-reason/spring/issues/16

TODO if you insist.

rts/Lua/LuaMatrixImpl.h
```diff
@@ -43,8 +43,9 @@ using tuple16f = std::tuple< float, float, float, float, float, float, float, fl
 class LuaMatrixImpl {
 public:
 	LuaMatrixImpl() = default; //matrices should default to identity
+	LuaMatrixImpl(LuaMatrixImpl&& lmi) = default; //move cons
```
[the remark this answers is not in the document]
lhog: It was more a cheat sheet for one C++ noob. :)

rts/builds/legacy/CMakeLists.txt
```diff
@@ -31,8 +31,9 @@ if    (FONTCONFIG_FOUND)
 	if    (PREFER_STATIC_LIBS)
 		prefer_static_libs()
 		find_library(EXPAT_LIBRARY expat)
+		find_library(LIBUUID_LIBRARY uuid)
```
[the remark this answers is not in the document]
lhog: I don't remember, but I couldn't link without it.

rts/System/SpringMath.h
```diff
@@ -119,6 +120,38 @@ template<class T> constexpr T Clamp(const T v, const T vmin, const T vmax) { ret
...
+template<typename TIn, typename TOut>
+TOut TransformFunc(const TIn input)
+{
+	if constexpr (std::is_same_v<TIn, TOut>)
+		return input;
+
+	constexpr TOut minOut = std::numeric_limits<TOut>::lowest();
+	constexpr TOut maxOut = std::numeric_limits<TOut>::max();
+
+	const TIn minIn = static_cast<TIn>(minOut);
```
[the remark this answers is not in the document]
lhog: Causes
`error: overflow in constant expression [-fpermissive]   constexpr TIn maxIn = static_cast<TIn>(maxOut);`

rts/System/SpringMem.h
```diff
@@ -0,0 +1,40 @@
...
+#include <map>
+
+namespace spring {
+	void* AllocateAlignedMemory(size_t size, size_t alignment);
+	void FreeAlignedMemory(void* ptr);
+}
+
+namespace spring {
+    template <typename T>
+    class StablePosAllocator {
```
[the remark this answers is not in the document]
lhog: Oh yes, it creeped in from one feature branch.

In essence the current implementation of MatrixUploader is less than ideal because the offset into SSBO array is not guaranteed to persist across draw frames, which makes it less usable than should from both engine (I have another half-done implementation that sources matrices for engine drawing completely from SSBO) and from Lua.

The future implementation (half done) will use this allocator to request a slot for as many unsynced matrices as needed.
For example for unit it will be world matrix + all piece matrices and will keep this slot until unit is dead. Upon death the slot will become free and either several smaller slots can be allocated in the free place or similar new unit can occupy it. The array from allocator will be uploaded to GPU's SSBO without any extra mess. Some stuff is done here: https://github.com/beyond-all-reason/spring/tree/BAR-matricesAllocator but I'm less than happy with overall implementation so it might get scrapped. The allocator itself proved to be robust in my "give all"/kill all and AI tests.

rts/Lua/LuaAtlasTextures.cpp
```diff
@@ -0,0 +1,104 @@
...
+{
+	for (auto& kv : textureAtlasMap)
+		spring::SafeDelete(kv.second);
+
+	textureAtlasMap.clear();
+}
+
+std::string LuaAtlasTextures::Create(const std::string& name, const int xsize, const int ysize, const int allocatorType)
+{
+	CTextureAtlas* atlas = new CTextureAtlas(allocatorType, xsize, ysize);
```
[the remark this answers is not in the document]
lhog: TODO

rts/Lua/LuaAtlasTextures.cpp
```diff
@@ -0,0 +1,104 @@
...
+	atlas->SetName(name);
+
+	textureAtlasMap[lastIndex] = atlas;
+
+	std::ostringstream ss;
+	ss << prefix << lastIndex;
+
+	++lastIndex;
+
+	return ss.str();
```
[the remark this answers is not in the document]
lhog: TODO

rts/Map/ReadMap.h
lhog: I know it's less than ideal.
Here is the thing. The previous implementation was incorrect. It only worked fine if minimal height were getting decreased and maximal height - increased during the course of the game. It's not always the case:

* Terraform
* Waterlevel gadget (terraform internally too)
* Smoothing

@Floris spotted values were incorrect and it was also evident that costly bumpwater was sometimes get rendered although no piece of terrain was below 0.

With the above in mind the only solution I came up with is to track all heights in the map and use its properties to update min max on occasions.

This change has been live for couple of months at least and yes it costs something, but I have no other solution.

rts/Rendering/Env/CubeMapHandler.cpp
lhog: Don't remember exact reason. Possibly got killed accidentally.
Shouldn't be relevant though. Face textures occupy each face to fit exactly, so wrapping shouldn't matter.

rts/Rendering/GL/VAO.h
lhog: Shaving off the cost where I can. What's done is done.

rts/Rendering/GL/VAO.cpp
lhog: It nicely fit into headers.
Without the .cpp file I couldn't reference it in CMakeLists. Is there some other way?

I can move the implementation if you prefer.

rts/Rendering/GlobalRendering.h
```diff
@@ -14,7 +14,7 @@ struct SDL_version;
 struct SDL_Window;
 typedef void* SDL_GLContext;
 
-struct CMatrix44f;
+class CMatrix44f;
```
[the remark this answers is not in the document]
lhog: TODO

rts/Rendering/GL/VAO.cpp
lhog: Done

rts/Map/ReadMap.h
rtri: Yeah I see the issue now, `currHeightBounds` only widens over time. I think an iterated update would be a good compromise here: on simframe `y`, scan heightmap row `y % num_rows`. Reset the working bounds to [+inf,-inf] when `y % n == 0`, keep a copy of the previous bounds while the iteration is ongoing.

rts/Sim/Units/CommandAI/MobileCAI.cpp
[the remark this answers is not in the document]
rtri: I have no objection to that, we just have to automate exporting the tickets (at least all the ones still open) to github.

rts/Lua/LuaUnsyncedCtrl.cpp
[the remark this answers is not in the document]
rtri: In this case yes, it collapses 5 lines to 1 and avoids a copy / no-op `SetState` call.

rts/System/SpringMath.h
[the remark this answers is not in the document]
rtri: as expected, this must already be actively used for downcasts

rts/builds/legacy/CMakeLists.txt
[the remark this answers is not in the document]
rtri: weird, but whatever

rts/System/FileSystem/ArchiveScanner.cpp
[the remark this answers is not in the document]
rtri: Nothing significant indeed, but no harm in being verbose with user-facing errors. I vote in favor of `throw content_error("Dependent archive \"" + lwrCaseName + "\" (resolved to \"" + resolvedName + "\") not found");`

rts/Sim/Units/CommandAI/MobileCAI.cpp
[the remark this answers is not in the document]
lhog: I'm not sure it's worth the effort.

Mantis is so polluted with mix of old news and issues no one is going to fix.
I feel it should be just made read-only in case someone will need to dig old some issue.
If something is pressing was left off, people will report it on github.

rts/Lua/LuaUnsyncedCtrl.cpp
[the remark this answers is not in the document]
lhog: I'm not sure SetState(GetState()) is no-op. Besides assigning params (no-op) it also calls Update(). We will lose Update() if we collapse it like that.

rts/Game/Game.cpp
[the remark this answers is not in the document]
lhog: Removed commented out things.

rts/Lua/LuaUnsyncedCtrl.cpp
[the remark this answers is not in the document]
rtri: Right, forgot `SetState` also called `Update` along the way. In that case I'd rather have `const bool retval = camHandler->SetState(hasState? ParseCamStateMap(L, 1): camHandler->GetState());` where `GetState()` is simply:

```
CCameraController::StateMap GetState() const {
    CCameraController::StateMap sm;
    GetState(sm);
    return sm;
}
```

rts/Map/ReadMap.h
lhog: So you think processing a HM row every sim frame is better than tracking updates ad-hoc?

lhog:
As far as I can tell for now only one big item left undecided/unfixed: heightmap ref map.
The rest are more or less cosmetic issues.

rts/Map/ReadMap.h
rtri: Yes, especially when large areas are updated. `std::map` introduces cache misses that are avoided with (and hurt much more than) a simple linear sweep.

lhog: Have a look https://github.com/spring/spring/pull/559/commits/985d21952633dc103acbb669e272aacfb1c50c85

rtri: Looks good (just some minor nits: `const int idxBeg` etc), but on a 1024x1024 heightmap this will read ~35K samples per frame which might be pushing it. Can easily be tuned later though.

lhog: I tried for_mt() approach on 20x20 map, so a chunk is around 54k items.
The results are pure disaster:

Even code like this:

```	for_mt(idxBeg, idxEnd, [this](const int idx) {
		const float h = (*heightMapSyncedPtr)[idx];
	});
```
consumes 7% of CPU time (I'm on 24 core machine). If I add atomics to get min/max it gets to 27%.
Do you know why for_mt is so costly?

Existing in-thread implementation consumes 0.2% compared.

rtri: Moving a task into the pool hits the allocator, one or more worker threads have to be woken up (possibly all of them for `for_mt`), the queue needs to be managed, etc. It is meant more for longer-running background jobs where startup latency can be ignored.

Using atomics opens up another can of worms regarding cache coherency, letting each thread calculate bounds locally (i.e. over its own HM chunk) and merging them at the end is 99% likely to be faster.

lhog: I was able to get slightly better times with integration and use of https://taskflow.github.io/taskflow/ParallelReduction.html#A2ParallelReduction , but for now it's probably not worth the bother.

rts/Lua/LuaAtlasTextures.cpp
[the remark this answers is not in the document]
lhog: > which you can do because every atlas stores its own name

Atlas `name != idStr` so update of textureAtlasMap upon .back() to deleted place position movement cannot be done nicely.
Are you sure you can't stand grandma-style C++? Because getting rid of it here proved tedious.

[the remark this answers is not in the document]
rtri: Shouldn't be that tedious, `textureAtlasMap` just needs to map from `id`'s to indices (rather than from names) and `CTextureAtlas` extended with a `std::string id` member. The second part can even be skipped if you change the `CreateTextureAtlas` API to not accept user-chosen names, which are mostly irrelevant anyway.

abma:
can you please cherry pick the changes which where made to .travis.yml, especially

https://github.com/spring/spring/commit/d667ad0cafc4a286070ac15d8ff5fe335a4b0ca9

lhog:
Sure it can be done here or afterwards.

But why bother with travis?

I heard in November they announced new pricing plan that made a lot of projects migrate away from them. https://travis-ci.org/ itself says we will be shutting down the former domain and continue under the new one. Moreover we have github actions. Right now they are launched on-demand to save free tier minutes, but with usual spring commits activity it can probably be run on commits and PRs too.

lhog:
BTW, please fork https://github.com/beyond-all-reason/spring-static-libs into spring or give me fork rights.
This is my reverse engineering work to build linux static libs necessary to build linux version of spring.

abma:
> But why bother with travis?

because it does compile test + unit tests on every commit.

abma:
> BTW, please fork https://github.com/beyond-all-reason/spring-static-libs into spring or give me fork rights.
> This is my reverse engineering work to build linux static libs necessary to build linux version of spring.

why not change it upstream? duplicating code should be really avoided.

lhog:
> because it does compile test + unit tests on every commit.

Github CI can absolutely do same. But whatever not worth the argument at the present time.
Is https://github.com/spring/spring/commit/d667ad0cafc4a286070ac15d8ff5fe335a4b0ca9 the only one needs cherry-picking. Perhaps I can just checkout the whole files from 105.0.1 if there are too many others?

> why not change it upstream? duplicating code should be really avoided.
Whatever your prefer. I was going to just follow the cargo cult and do like others did for their side work (example: https://github.com/spring/CircuitAI ), i.e. to fork to spring and continue here.
That said I cannot do forks nor create new repos (insufficient set of rights). Thus my ask to give rights or do stuff yourself.

abma:
> Github CI can absolutely do same. But whatever... not worth the argument at the present time.
> Is [d667ad0](https://github.com/spring/spring/commit/d667ad0cafc4a286070ac15d8ff5fe335a4b0ca9) the only one needs cherry-picking? Perhaps I can just checkout the whole files from 105.0.1 if there are too many others?

i have no clue as it fails at a very early stage and have no clue how large the diff between develop and "transition". just please cherry-pick the commit to reduce the diff and very likely fix travis compile.

lhog:
Travis CI doesn't work on pure spring/spring/transition: See https://github.com/spring/spring/pull/560

lhog:
Ok travis doesn't seem to work with this PR and it probably it won't work w/o adaptation to the changes I did to make it build-able on github CI (and in Docker container).

abma:
https://travis-ci.org/github/spring/spring/jobs/771462288#L867 ->
this is what fails: https://github.com/spring/spring/pull/559/files#diff-48b508d8a5702095973f36f384be2a8b74834092bb7ed47c536f1af80015b20bR4

seems the check is invalid: https://cmake.org/cmake/help/latest/policy/CMP0077.html
CMP0077 requires cmake >= 3.12 and travis seems to use  3.12.4: not sure why "NOT CMAKE_VERSION VERSION_LESS 3.12" is used, i would use 
if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.12)

lhog:
Ok this is above my paygrade so to speak. Glad you know stuff.

Does travis have no cmake replacement procedure? Like here: https://github.com/beyond-all-reason/spring/blob/BAR/.github/workflows/build.yml#L98-L101 ?

We will also need newer GCC. Something like 9.x or better 10.x.

rts/Map/ReadMap.cpp
```diff
@@ -568,13 +568,6 @@ void CReadMap::UpdateHeightBounds(int syncFrame)
 		tempHeightBounds.y = std::max(h, tempHeightBounds.y);
 	}
 #else
```
rtri, line 570: The complexity / performance tradeoff is not favorable here. This function now performs an order of magnitude more work for a result that is identical within measurement error to the ST version, and is much longer to read and harder to understand at a glance. Moreover, two obvious optimizations have not been considered: an early-out dirty check if the heightmap was changed at all and use of SSE intrinsics (`minps`, don't ever trust the compiler to emit these for you) to vectorize the loop.

The code can stay as a reference if desired, but please place it within the `#if 0` block.

rts/Map/ReadMap.cpp
```diff
@@ -559,13 +559,64 @@ void CReadMap::UpdateHeightBounds(int syncFrame)
...
+	int chunksPerThread = 1;
+	int usingThreads = minChunksForCacheLine;
+
+	if (minChunksForCacheLine > maxThreads) {
+		chunksPerThread = minChunksForCacheLine / maxThreads;
+		chunksPerThread = minChunksForCacheLine % maxThreads ? chunksPerThread + 1: chunksPerThread;
+		usingThreads = maxThreads;
+	}
+
+	std::vector<float2> threadResults(usingThreads);
```
rtri, line 586: should be `std::array<float2, ThreadPool::MAX_THREADS>`, this can easily live on the stack

rts/Map/ReadMap.cpp
lhog: Done. Employed interesting lib for that. The code is more readable than pure SSE intrinsics

rtri: Nice, but does it sync? What controls which SIMD instruction set(s) this library uses?

lhog: Should sync. Emitted assembly uses instructions {min,max}{ps,ss}.
Though all recent changes, requested during the review of this PR, are not battle tested.

The level SIMD instructions is defined by conditional compilation https://xsimd.readthedocs.io/en/latest/api/instr_macros.html and compilers are supposed to automatically define definitions like `__SSE__`. At least this is what stackoverflow says. In worst case scenario `XSIMD_X86_INSTR_SET` to appropriate `XSIMD_X86_*_VERSION` manually through CMakeLists.

rts/Lua/LuaAtlasTextures.cpp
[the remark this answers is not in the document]
lhog: Have a look: https://github.com/spring/spring/pull/559/commits/57db9665e4916480031aa716c848ccbe30beee78

rts/Lua/LuaConstPlatform.cpp
```diff
@@ -30,7 +30,9 @@ bool LuaConstPlatform::PushEntries(lua_State* L)
 	LuaPushNamedBool(L, "glSupportNonPowerOfTwoTex", globalRendering->supportNonPowerOfTwoTex);
 	LuaPushNamedBool(L, "glSupportTextureQueryLOD" , globalRendering->supportTextureQueryLOD);
 	LuaPushNamedBool(L, "glSupportMSAAFrameBuffer" , globalRendering->supportMSAAFrameBuffer);
-	LuaPushNamedBool(L, "glSupport24bitDepthBuffer", globalRendering->support24bitDepthBuffer);
+
+	LuaPushNamedBool(L, "supportDepthBufferBitDepth", globalRendering->supportDepthBufferBitDepth);
```
rtri, line 34: should be `LuaPushNamedNumber` now

rts/Map/ReadMap.h
```diff
@@ -286,16 +291,18 @@ extern MapDimensions mapDims;
...
 inline float CReadMap::AddHeight(const int idx, const float a) { return SetHeight(idx, a, 1); }
 inline float CReadMap::SetHeight(const int idx, const float h, const int add) {
-	float& x = (*heightMapSyncedPtr)[idx];
+	float& heightRef = (*heightMapSyncedPtr)[idx];
 
 	// add=0 <--> x = x*0 + h =   h
 	// add=1 <--> x = x*1 + h = x+h
-	x = x * add + h;
+	float newHeight = heightRef * add + h;
+	if (newHeight == heightRef)
```
rtri, line 299:
```
updateHeightBounds |= (newHeight != heightRef);
return (heightRef = newHeight);
```

or

`return (updateHeightBounds |= (newHeight != heightRef), heightRef = newHeight);`

lhog: I personally think such style  (`return (x = y);` or `return (x, y);`) hurts readability, but I guess it's matter of preference.
Committed 1st option.

rts/Rendering/GL/VBO.h
```diff
@@ -52,13 +52,19 @@ class VBO
 	 * @see http://www.opengl.org/sdk/docs/man/xhtml/glBufferData.xml
 	 */
 	void Resize(GLsizeiptr newSize, GLenum newUsage = GL_STREAM_DRAW);
+
+	template<typename TData>
+	void New(std::vector<TData> data, GLenum newUsage = GL_STATIC_DRAW) { New(sizeof(TData) * data.size(), newUsage, data.data()); };
```
rtri, line 57: `const std::vector<TData>& data`, no copies please

rts/Rendering/GL/VBO.h
```diff
@@ -52,13 +52,19 @@ class VBO
...
 	void New(GLsizeiptr newSize, GLenum newUsage = GL_STREAM_DRAW, const void* newData = nullptr);
+
 	void Invalidate(); //< discards all current data (frees the memory w/o resizing)
 
 	/**
 	 * @see http://www.opengl.org/sdk/docs/man/xhtml/glMapBufferRange.xml
 	 */
-	GLubyte* MapBuffer(GLbitfield access = GL_WRITE_ONLY);
+	template<typename TData>
+	GLubyte* MapBuffer(std::vector<TData> data, GLintptr elemOffset = 0, GLbitfield access = GL_WRITE_ONLY) { return MapBuffer(sizeof(TData) * elemOffset, sizeof(TData) * data.size(), access); };
```
rtri, line 66: `const std::vector<TData>& data`

rts/Rendering/GL/VBO.h
```diff
@@ -70,10 +76,10 @@ class VBO
 		UnmapBuffer();
 		Unbind();
 	}
-	void SetBufferSubData(GLintptr offset, GLsizeiptr size, void* data);
-
-
 
+	template<typename TData>
+	void SetBufferSubData(std::vector<TData> data, GLintptr elemOffset = 0) { SetBufferSubData(sizeof(TData) * elemOffset, sizeof(TData) * data.size(), data.data()); }
```
rtri, line 81: `const std::vector<TData>& data`

rts/Rendering/Fonts/CFontTexture.cpp
```diff
@@ -531,16 +531,16 @@ void CFontTexture::LoadBlock(char32_t start, char32_t end)
 	do {
 		alreadyCheckedFonts.insert(f);
 
-		for (auto it = map.begin(); !map.empty() && it != map.end(); ) {
-			FT_UInt index = FT_Get_Char_Index(*f, *it);
+		for (std::size_t idx = 0; idx < map.size() && !map.empty(); /*nop*/) {
```
rtri, line 534: No idea why `!map.empty()` was present before, but `idx < map.size()` should be a sufficient condition.

rts/Rendering/Textures/TextureAtlas.h
```diff
@@ -40,9 +38,27 @@ class CTextureAtlas
 	};
 
 public:
-	CTextureAtlas(unsigned int allocType = ATLAS_ALLOC_LEGACY);
+	CTextureAtlas(unsigned int allocType = ATLAS_ALLOC_LEGACY, const int atlasSizeX_ = 0, const int atlasSizeY_ = 0, std::string name_ = "");
+
+	CTextureAtlas(CTextureAtlas&& ta) noexcept
+	{
+		MoveFields(std::forward<CTextureAtlas>(ta));
```
rtri, line 45: The usual trick for move-constructors is to write `*this = std::move(ta);`. That way you don't need a separate `MoveFields`, just the move-assignment operator.

abma:
can you please readd the AIs + fix the tests? for the failing ais see
https://travis-ci.org/github/spring/spring/jobs/771996375

+ the linking error https://travis-ci.org/github/spring/spring/jobs/771996376:
[ 91%] Linking CXX executable test_Matrix44f
CMakeFiles/test_Matrix44f.dir/__/rts/System/Matrix44f.cpp.o:Matrix44f.cpp:function CMatrix44f::ClipOrthoProj01(): error: undefined reference to 'globalRendering'
collect2: error: ld returned 1 exit status
test/CMakeFiles/test_Matrix44f.dir/build.make:353: recipe for target 'test/test_Matrix44f' failed
make[3]: *** [test/test_Matrix44f] Error 1
make[3]: Target 'test/CMakeFiles/test_Matrix44f.dir/build' not remade because of errors.
CMakeFiles/Makefile2:5189: recipe for target 'test/CMakeFiles/test_Matrix44f.dir/all' failed
make[2]: *** [test/CMakeFiles/test_Matrix44f.dir/all] Error 2

lhog:
> can you please readd the AIs? for the failing ais see

Generally yes, but I think we need to do certain things:
* Remove ancient AIs: HughAI, some others no one uses()
* Add new BARbarian, modified a CircuitAI that works with BAR
* Agree how to rewrite the part of builder script that I'm least proud of: https://github.com/beyond-all-reason/spring/blob/BAR/.github/workflows/build.yml#L422-L433

> fix the tests, linking fails https://travis-ci.org/github/spring/spring/jobs/771996376:

Sure. In the tasks queue.

abma:
@rtri:

what do you think about renaming the branch "transition" to "develop" and "develop" to "opengl4-test"?

(as renaming seems not possbile, it would be a revert + merge commit in the develop branch)

we really need to get rid of two development branches.

abma:
> * Remove ancient AIs: HughAI, some others(if any) no one uses

i don't care much about which ai's are kept: but the tests needs to be fixed.

> Agree how to rewrite the part of builder script that I'm least proud of: https://github.com/beyond-all-reason/spring/blob/BAR/.github/workflows/build.yml#L422-L433

for now use "make install" instead of manually coping files and hard-coding a lot of paths.

rts/lib/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ INCLUDE_DIRECTORIES(${CMAKE_SOURCE_DIR}/rts)
 
 ADD_SUBDIRECTORY(lua)
 ADD_SUBDIRECTORY(luasocket)
+ADD_SUBDIRECTORY(meshoptimizer)
```
abma, line 6: side node: some additions in rts/lib seems to not contain the "LICENSE" files, i.e. https://github.com/spring/spring/blob/d4eefe88a865d41f4f25f3db072d1cad751728a7/rts/lib/xsimd/xsimd.hpp

when possible IMHO it would be better to use git submodules: this should make it easier to keep dependencies up to date.

thats maybe something which should be solved when we have a conclusion on about develop / transition branch changes + the roadmap.

abma (changes requested):
-> the license files should be (re-)added

lhog:
> i don't care much about which ai's are kept: but the tests needs to be fixed.

Ok, I'll have a look some time later

> for now use "make install" instead of manually coping files and hard-coding a lot of paths.

Last time I checked `make install` did not produce the operational files/directory structure, that one could put into zip and distribute as minimal standalone engine package. Can you work on `make install` or say `make distro` script so it copies all files into the install dir such that it can be packaged as zip later on?

> side node: some additions in rts/lib seems to not contain the "LICENSE" files

Yeah it's accidental. I'll restore.

> when possible IMHO it would be better to use git submodules: this should make it easier to keep dependencies up to date.

Not sure it's the for the best. Some libs have "fat" like tests, benchmarks, examples, additional libs, etc. trimmed off.

abma:
> Can you work on make install or say make distro script so it copies all files into the install dir such that it can be packaged as zip later on?

done: https://github.com/spring/spring/blob/develop/buildbot/slave/linux/create_linux_static_bundle.sh

lhog:
I'm yet to find out how you dealt with AI versions/tags. Yet in principle it is only marginally to what github builder does. I was talking about make command implementation, which everyone in Nix world are used to.

abma:
> I'm yet to find out how you dealt with AI versions/tags. But in principle it is only marginally better to what github builder does. I was talking about make command implementation, which everyone in Nix world are used to.

i don't understand what you mean. the script uses "make install": https://github.com/spring/spring/blob/develop/buildbot/slave/linux/create_linux_static_bundle.sh#L19

the only thing which happens in the .sh is stripping + creating the .7z archives.

lhog:
> i don't understand what you mean. the script uses "make install": https://github.com/spring/spring/blob/develop/buildbot/slave/linux/create_linux_static_bundle.sh#L19

Ah I missed that. Makes sense. I'll give it a try again.

lhog:
@abma  I fixed tests compilation, AI tests and test results.

abma:
> @abma I fixed tests compilation, AI tests and test results.

nice, thanks!

lhog:
@abma I rechecked `make install`. Here are my findings:
* if `AI_TYPES` = `NONE`, then no AI dir is created by install. Is this expected? I'd expect it to create AI dir and put interfaces there, but not the actual skirmishers.
* Files structure is a bit different from what I get from buildbot:

```
ls -al /spring/build-linux-64/install/
total 940488
drwxr-xr-x 10 root root      4096 May 28 02:24 .
drwxr-xr-x 12 root root      4096 May 28 02:29 ..
drwxr-xr-x  4 root root      4096 May 28 02:24 AI
drwxr-xr-x  9 root root      4096 May 28 02:21 LuaUI
drwxr-xr-x  3 root root      4096 May 28 02:21 base
-rw-r--r--  1 root root      2308 Jan 28 13:40 cmdcolors.txt
-rw-r--r--  1 root root      2757 Jan 28 13:40 ctrlpanel.txt
drwxr-xr-x  3 root root      4096 May 28 02:21 examples
drwxr-xr-x  2 root root      4096 May 28 02:21 fonts
drwxr-xr-x  3 root root      4096 May 28 02:21 include
drwxr-xr-x  4 root root      4096 May 28 02:21 lib
-rw-r--r--  1 root root  49744456 May 28 02:18 libunitsync.so
-rw-r--r--  1 root root      1822 Jan 28 13:40 luaui.lua
-rwxr-xr-x  1 root root   9010144 May 28 02:18 mapcompile
-rwxr-xr-x  1 root root   8911312 May 28 02:18 mapdecompile
-rwxr-xr-x  1 root root  16018080 May 28 02:18 pr-downloader
drwxr-xr-x  5 root root      4096 May 28 02:21 share
-rw-r--r--  1 root root      3987 Jan 28 13:40 socket.lua
-rwxr-xr-x  1 root root 415526072 May 28 02:20 spring
-rwxr-xr-x  1 root root  62483160 May 28 02:20 spring-dedicated
-rwxr-xr-x  1 root root 401292680 May 28 02:21 spring-headless
```
I have redundant `include` and `lib`:

```
ls -al /spring/build-linux-64/install/include/
total 72
drwxr-xr-x  3 root root  4096 May 28 02:34 .
drwxr-xr-x 10 root root  4096 May 28 02:34 ..
drwxr-xr-x  3 root root  4096 May 28 02:34 git2
-rw-r--r--  1 root root  1803 Jan 28 13:40 git2.h
-rw-r--r--  1 root root 54819 May 28 02:04 meshoptimizer.h

ls -al /spring/build-linux-64/install/lib/
total 14776
drwxr-xr-x  4 root root     4096 May 28 02:34 .
drwxr-xr-x 10 root root     4096 May 28 02:34 ..
drwxr-xr-x  3 root root     4096 May 28 02:34 cmake
-rw-r--r--  1 root root 13579140 May 28 02:18 libgit2.a
-rw-r--r--  1 root root  1529636 May 28 02:17 libmeshoptimizer.a
drwxr-xr-x  2 root root     4096 May 28 02:34 pkgconfig
```
Do you know where they might come from?

This step probably makes sense https://github.com/spring/spring/blob/develop/buildbot/slave/linux/create_linux_static_bundle.sh#L57 to port over to `make install` too.

All in all include & lib aside seems like I can replace my github script crap mostly with just `make install`

abma:
> if AI_TYPES = NONE, then no AI dir is created by install. Is this expected? I'd expect it to create AI dir and put interfaces there, but not the actual skirmishers.

afaik only the ai interfaces are build which are required for the requests ai's so yes.

> Do you know where they might come from?

differnt cmake flags. these files shouldn't be installed, thats a bug in the CMakeLists.txt. libgit stuff comes from pr-downloader, 
libmeshoptimizer.a is very likely there because of "install()" here: https://github.com/spring/spring/blob/382db9ff77fb842bab9d9ba2175442100ed652b1/rts/lib/meshoptimizer/CMakeLists.txt#L68

this maybe is partly fixed in "develop".

lhog:
@abma nice! Thanks for enlightening me. I can probably fix meshoptimizer part, but I'm not sure how to deal with libgit, given it's submodule.

lhog:
@rtri do you have any time estimate on when you expect to finish the review?

rtri:
@lhog no ETA but I'll continue this weekend.

lhog:
@rtri no problem, take as much time as needed.
For information I'm piling up additional commits here: https://github.com/beyond-all-reason/spring/pull/35

.gitmodules
```diff
@@ -17,3 +14,7 @@
 	path = AI/Skirmish/BARb
 	url = https://github.com/rlcevg/CircuitAI.git
 	branch = barbarian
+[submodule "AI/Skirmish/CircuitAI"]
+	path = AI/Skirmish/CircuitAI
+	url = https://github.com/rlcevg/CircuitAI.git
```
abma, line 19: please revert this change. we had it several times that required submodules disappeared because of renames or deletion. thats why all submodules are hosted at https://github.com/spring/*

abma, line 19: please revert this change. we had it several times that required submodules disappeared because of renames or deletion. thats why all submodules are hosted at https://github.com/spring/*

lhog: Talk to @rlcevg (lamer). He is the author of both AIs and this PR, I suppose he knows better.

P.S. @abma check your springrts PMs, there is an emergency waiting your attention for a while now.

P.S.S I'm on PTO, but once I'm back I'm gonna start putting new commits to this branch as the review has stalled.

rlcevg: How the heck BARbarIAn and CircuitAI that will use new or changed API functions (but not yet) ended up in spring repo?
Is Beyond-All-Reason/spring/BAR going to be merged completely into spring/spring/transition? (if so then reverting is fine)

lhog: Ok, it is not outside of my control that this is taking so long. Had I knew this beforehand, I'd have made this branch separate from BAR live one.

Let's agree on the following: once the lamer's changes to AI have made it to the spring/spring (as part of this branch), we will certainly revert submodules URL back.

lhog: Alternatively once I'm back we can craft a separate spring/spring AI PR and I can quickly merge it to BAR too. This way submodules change won't be needed.

lhog: Alternatively once I'm back we can craft a separate spring/spring AI PR and I can quickly merge it to BAR too. This way submodules change won't be needed.

rlcevg: dependency for submodule revert: https://github.com/spring/CircuitAI/pull/19

abma:
i can only speak for me but as it is, its impossible to review this pull request as it basicly touches all places of the engine and seems to not come to a "stable" final state, its way to large and i can understand why the motivation was lost at some point to review it.

as it currently is, it seems you successfully created a fork.

ashdnazg:
I think the mistake with this PR is that it's a moving target - this should have been done in small parts each from a snapshot of BAR.

lhog:
First of all @abma @ashdnazg the latest merge should've gone into the feature branch, if needed I can revert or reset to the previous state. We'll sort it out no worries.

BAR branch became a fork long ago due to obvious stagnation of the primary branch, yet I've put (and still putting) every effort so we merge back.

That said I warned we cannot hold ourselves back much longer. The review has been going for 2.5 months and I implemented every change requested. All that time for BAR branch was mostly receiving critical fixes and small QoL improvements . The need to not push too much on BAR branch slows us down: I personally develop the continuation of GL4 work elsewhere and I have a lot of operational issues because of that.

If the merge is still desired from spring side, let's decide something on the branches situation. Right now it's quite messy:
* `develop` is unused and no one is planning to use it. Yet I see new commits are being occasionally pushed there. 
* `master` seems to be used to hold the state of the latest stable, which is not what most people use master for.
* `transition` (bad name by now) is targeted as a future home for what known as BAR branch

Can we agree
* to archive `develop` and salvage what possible (quite a lot) from there?
* rename `transition` to `master`
* abolish former `master` as release tags render it rather useless

...and finally develop one engine in one branch and not 3 of them?

RFC: @gajop @rtri @ashdnazg @abma

ashdnazg:
> the latest merge should've gone into the feature branch, if needed I can revert or reset to the previous state
I think there were more commits and PRs that weren't just fixes of things mentioned here, but I might be wrong

> The need to not push too much on BAR branch
This should never have been the case - you should have had a specific branch for this PR and continued to push other stuff to the main BAR branch.

> let's decide something on the branches situation
Tags are one thing and master/develop is another. The convention oif holding the latest stable on master is not something we've made up.
If you suggest dropping the current develop, it would be renamed to something else and transition would become the new develop.

In any case, I cannot commit to reviewing this PR in any reasonable timeframe, especially when it's expanding all the time, so as far as I can see it either:
1) somebody else can do it
2) ivand gets jurisdiction over the spring repo
3) the PR can be closed as there's little hope for convergence.

lhog:
Ok, the PF commits are now gone (sorry you'll likely need to pull --rebase this one).
https://github.com/beyond-all-reason/spring/pull/55 now has the valuable TKPF work by Marcus / TK.

badosu:
Any chance of this ever being merged? I feel failure will probably cause a permanent divergence, both technically and in terms of motivations.

lhog:
Sorry, need to break feature freeze with desync fix (and minor update happened to be waiting in the local stash).

lhog:
Another fix for crashes  :|
Thanks @ashdnazg !

lhog:
Greetings.

As I didn't expect the review to take long it didn't come to my mind to separate life branch from review branch and now it's troublesome.

With that inconvenience in mind I tried to defer adding low priority items into BAR branch as much as I could. But there will be an exception soon. BAR artists asked for a few CEG improvements that I did (as best as I could). These improvements are already committed into the game and wait for the engine update that would enable them. Therefore, soon I'll be merging these CEG commits.

rts/lib/lua/src/lvm.cpp
```diff
@@ -10,6 +10,7 @@
 #include <string.h>
 
 //SPRING
+#include <algorithm>
```
rtri, line 13: for peace of mind, three more things should be done regarding this desync:

1) LC_NUMERIC should be changed to LC_ALL in https://github.com/spring/spring/blob/develop/rts/System/Main.cpp#L37
2) access to `os.setlocale` should be removed entirely (let https://github.com/spring/spring/blob/develop/rts/lib/lua/src/loslib.cpp#L231 `return 0`)
3) the locale should be net-logged somewhere on startup, for example at https://github.com/spring/spring/blob/develop/rts/Game/Game.cpp#L437 (change to `SNPRINTF(msgBuf, sizeof(msgBuf), "[Game::%s][lua{Rules,Gaia}={%p,%p}][locale=%s]", __func__, luaRules, luaGaia, setlocale(LC_ALL, nullptr));`)

rts/Sim/Misc/Team.cpp
```diff
@@ -349,8 +349,15 @@ void CTeam::SlowUpdate()
 			if (team->isDead)
 				continue;
 
-			const float edif = std::max(0.0f, (team->resStorage.energy * 0.99f) - team->res.energy) * de;
```
rtri, line 352: `const float edif = Clamp(((team->resStorage.energy * 0.99f) - team->res.energy) * de, 0.0f, res.energy);`
`const float mdif = Clamp(((team->resStorage.metal * 0.99f) - team->res.metal) * dm, 0.0f, res.metal);`

Don't keep the old lines.

rts/Sim/Misc/Team.cpp
```diff
@@ -349,8 +349,15 @@ void CTeam::SlowUpdate()
...
-			const float mdif = std::max(0.0f, (team->resStorage.metal * 0.99f) - team->res.metal) * dm;
+			//due to precision errors mdif/edif sometimes can be slightly >= than res. If team has no metal income
+			//this causes units with zero fire resources requirements to unable to fire
+			//due to CTeam::HaveResources() failed check
```
rtri, line 354: `this causes units with zero fire resources requirements to be unable to fire`
`when CTeam::HaveResources() is evaluated, thus clamp edif/mdif on both sides`

rts/Game/InMapDrawModel.cpp
```diff
@@ -161,24 +161,28 @@ void CInMapDrawModel::EraseNear(const float3& constPos, int playerID)
 		for (int x = xStart; x <= xEnd; ++x) {
 			DrawQuad* dq = &drawQuads[(y * drawQuadsX) + x];
 
-			for (auto pi = dq->points.begin(); pi != dq->points.end(); /* none */) {
+			//for (auto pi = dq->points.begin(); pi != dq->points.end(); /* none */) {
```
rtri, line 164: No point in commenting out the old loops if you don't leave behind a comment explaining why you chose explicit indexing, tell the reader MSVC chokes on iterator manipulation.

rts/System/GlobalRNG.h
```diff
@@ -128,8 +128,8 @@ template<typename RNG, bool synced> class CGlobalRNG {
 	rng_res_type NextInt(rng_res_type N = max()) { return ((*this)(N)); }
 
 	float NextFloat() { return (NextFloat01(1 << ndig())); }
-	float NextFloat01(rng_res_type N) { return ((NextInt(N) * 1.0f) / N); } // [0,1) rounded to multiple of 1/N
-	float NextFloat24() { return (math::ldexp(NextInt(1 << ndig()), -ndig())); } // [0,1) rounded to multiple of 1/(2^#digits)
+	float NextFloat01(rng_res_type N) { return static_cast<float>(NextInt(N)) / N; } // [0,1) rounded to multiple of 1/N
```
rtri, line 131: 1) multiplying `T` by `1.0f` promotes `T` to a float, which has the same effect as `static_cast`'ing it (but is shorter)
2) similarly `ldexp` implicitly converts its argument to a float (`streflop::Simple` technically), why the cast?
3) not to be blunt, but why are you touching RNG code in the middle of a CEG-related commit bundle?

rtri:
FYI I'm slowly making my way through https://github.com/spring/spring/pull/559/commits/cd328de80f359f47c37d4a19247ee1c9ae6cbc4e before returning to the main part of this PR.

I sincerely do hope the finish line stops moving now.

rts/Game/InMapDrawModel.cpp
lhog: Deal, will be fixed.

rts/Sim/Misc/Team.cpp
lhog: ok

rts/Sim/Misc/Team.cpp
lhog: ok, will change.

rts/lib/lua/src/lvm.cpp
lhog: Sure, will do

rts/Sim/Projectiles/ProjectileMemPool.h
```diff
@@ -5,11 +5,16 @@
 
 #include "Sim/Misc/GlobalConstants.h"
 #include "System/MemPoolTypes.h"
+#include "System/SpringMath.h"
+
+#include "Sim/Projectiles/WeaponProjectiles/StarburstProjectile.h"
+
+static constexpr size_t PMP_S = AlignUp(sizeof(CStarburstProjectile), 4); //biggest in size
```
rtri, line 12: The magic constant was here to avoid including any specific projectile type header. If some other projectile type ever becomes the largest footprint you'll have to change both the `#include` and the `PMP_S` LOC so this really adds nothing over just keeping a raw number around.
A clean solution would be calculating max(sizeof(T1), sizeof(T1), ...) at compile-time, which also requires including every single projectile type header and just isn't worth it.

lhog: Respectfully disagree. The day when some projectile will suddenly outgrow Starburst may never come, while the need to do guesswork of how big the magic constant should be will come right after a new field is added to CProjectile (or its base classes).

rts/System/ContainerUtil.h
```diff
@@ -102,7 +103,16 @@ namespace spring {
...
 
+	//do sort according to predicate p after every emplace_back of t to v
+	template<typename T, typename P = std::less<>>
+	static void VectorPushBackSorted(std::vector<T>& v, const T& t, P&& p = P()) {
+		v.emplace_back(t);
 
+		const std::size_t n = v.size() - 1;
+
+		//std::sort(std::begin(v) + n, std::end(v), p);
+		std::inplace_merge(std::begin(v), std::begin(v) + n, std::end(v), p);
```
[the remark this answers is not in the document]
lhog: Yep I screwed up. Fix is incoming.

sprunk:
See https://github.com/beyond-all-reason/spring/commit/bf17d0b63dc7be529fb508e5a3509acd6eaceb70 about the reasoning for the last commit.

lhog:
Guys, I feel like half a year was enough time, yet the review is not complete nor moving forward.

The need to carry tons of legacy, BAR doesn't need, and maintain 100% compatibility with 105 spring (BAR doesn't need either) is a heavy burden. I don't want to carry it pointlessly if for some undisclosed reason the merge back is not desired.

With that in mind my offer to merge this branch back expires on the 1st of November.

sprunk:
> So, what's delaying the merge?

Somebody needs to review the PR. @rtri was halfway through but there's a bit of a silence about what's going on.

lhog:
All right.

It's 1st of November and as announced I'm closing this PR due to no tangible review progress and the fact that this(BAR) branch is now obsolete and unmaintained. The development has moved on to https://github.com/beyond-all-reason/spring/tree/BAR105 where features and fixes are added almost daily.

If someone ever wants to continue review they should check out BAR105 branch, but asynchronously (no commit/feature freezes on the branch will be done).

Nevertheless, I'd like to thank all reviewers for their valuable advises, you certainly made the code much better.

tgunnoe:
Sad that this had to happen

PR #562, opened 2021-05-17
Fix depth check for non-fac seaplane buildability

sprunk:
Before the patch:
 * neither `canSubmerge` nor `floater`: NOT buildable on water
 * just `canSubmerge`, but not `floater`: buildable on water
 * just `floater`, but not `canSubmerge`: NOT buildable on water
 * both `floater` and `canSubmerge`: NOT buildable on water

After the patch:
 * neither `canSubmerge` nor `floater`: NOT buildable on water
 * just `canSubmerge`, but not `floater`: buildable on water
 * just `floater`, but not `canSubmerge`: buildable on water
 * both `floater` and `canSubmerge`: buildable on water


lhog:
Judging from the truth table seems to make sense

PR #563, opened 2021-06-01
All whispers are visible in a replay

sprunk:


ashdnazg:
I'm not sure I like this as a default - you shouldn't have access to a conversation between two people without being certain they know about it...

ashdnazg:
I never knew whispers are even recorded in a demo. In any case, this won't help anything except making people speccheat in other ways.

sprunk:
Who said it's just for spec cheating? Maybe I would like to learn how to play FFA by watching a high skill replay and seeing how pros ~~lie to each other~~ conduct diplomacy would be helpful.

I consider the current state of affairs a bug, because I think everything in a replay should be visible unless there is a good reason not to. I don't consider privacy a good reason in this case because whispers are already in the demo so anybody dedicated can extract them anyway. Possibly whispers shouldn't be recorded in the demo instead, but I think that would go against the point of a replay.

lhog:
Please don't forget to cherry-pick into transition.

abma:
as its stored in the replay file its IMHO correct to show it.

PR #565, opened 2021-06-25
Transition changelog: add 106, move recent stuff thither

sprunk:


lhog (approved):
Yeah, we need to starting filling it one by one.

PR #568, opened 2021-12-26
[unitsync] Default install destination

mfld-fr:
The `libunitsync.so` is force installed in `/usr/lib`, but depending on the architecture and the target, that folder can be `lib`, `lib32`, `lib64`... so let the default destination in `install()`, thus allowing the CMake caller to specify the right destination folder through `CMAKE_INSTALL_LIBDIR` (see https://cmake.org/cmake/help/v3.22/command/install.html).

I am not enough experienced with Spring RTS to propagate that change to the whole source tree, so only modified the `CMakeLists.txt` that impacts the actual destination of `libunitsync.so`. I suggest to cleanup the whole `LIBDIR` stuff accordingly.

### DIFF
--- a/tools/unitsync/CMakeLists.txt
+++ b/tools/unitsync/CMakeLists.txt
@@ -147,9 +147,9 @@ if (MINGW)
 endif (MINGW)
 fix_lib_name(unitsync)
 
-install (TARGETS unitsync
-	RUNTIME DESTINATION ${LIBDIR}
-	LIBRARY DESTINATION ${LIBDIR})
+include(GNUInstallDirs)
+install(TARGETS unitsync
+	LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
 
 add_subdirectory(test)
 


PicassoCT:
Ivand on BAR:
> I'm not entirely sure what the point of system wide installation is?
> Is it done to run all games on a single engine version be it 105 or 106 or w/e? If so - this will never work for us as we switch engines once a week or so.

mfld-fr:
Hello @PicassoCT, thanks for the intent to wake up engine devs, but I would like to make the things clear on my POV & interest. I came to Spring because of S44 and its historical background, not because of any other futuristic games, whatever their quality and that very special pleasure to watch firework on battlefield with many FX :laughing:

I see no reason to involve the BAR community, that looks like (as far as I understand) to have no more interest in maintaining the original engine, and is focusing on their full fork (engine + game + infra). I have no opinion about that forking, that's life in open-source, and I wish the best success for them.

Back to the proposed patch, it has nothing to do with a system wide install, as the CMake caller can specify the final location with the [CMAKE_INSTALL_PREFIX](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html) variable (or `DESTDIR`). And it has nothing to do either with any kind of 'folder autodetection'.

Let us just wait for @abma to come in and push the green button...

mfld-fr:
@abma : this one targets the `develop` branch :smirk:

abma:
IMHO thats the wrong place, if changed, it should be changed globally:

https://github.com/spring/spring/blob/develop/CMakeLists.txt#L106

can't you set LIBDIR in the build script? i.e. if i remember right:

 cmake . -DLIBDIR=...

abma:
thats how debian changes the path:

mfld-fr:
@abma : Respectfully disagree : it looks like distros are looking to cleanly separate the 32 / 64-bit libraries while trying to keep maximum compatibility with legacy 32-bit installations for multi-lib environment. As stated in https://www.gentoo.org/support/news-items/2019-06-05-amd64-17-1-profiles-are-now-stable.html:
```
In the new profiles, the lib->lib64 compatibility symlink is removed.
64-bit libraries need to be installed directly to lib64.  /lib
and /usr/lib become real directories, that are used for cross-arch
and native non-library packages (gcc, clang) and 32-bit libraries
on the multilib profile (which improves compatibility with prebuilt x86
packages).
```
I think the global variable `LIBDIR` should be kept and still used to install Spring stuff that is 32-bit or cross-arch in `lib`, but the 64-bit `libunitsync.so` should go to `lib64`, so the local change for that library only in that PR.

abma:
afaik there is nothing cross-arch which should be in "lib". its either lib32 or lib64...

so when compiled for 32bit use sth. like 

   cmake . -DLIBDIR=lib32

and for x64 use

   cmake . -DLIBDIR=lib64

mfld-fr:
@abma : as you want, but I still think there is no reason to force `LIBDIR` on the CMake command line while it automatically defines `CMAKE_INSTALL_LIBDIR` to the right value, even for Windows. Taking into account your last sentence that learn me `libunitsync.so` is the unique library, the global change should be then to keep my local change for that library and cleanup the global and useless `LIBDIR`.

abma:
ok, then i missread what the goal of this change is:

you want to change the default install locations to match the GNUInstallDirs.

-> https://github.com/spring/spring/blob/develop/CMakeLists.txt#L105

these lines should be changed to:

  set(BINDIR  "${CMAKE_INSTALL_BINDIR}"                CACHE STRING "Where to install binaries")
  set(LIBDIR  "${CMAKE_INSTALL_LIBDIR}"                CACHE STRING "Where to install libraries")
[...]

PR #570, opened 2022-03-16
Revert 980dda9bf26d3d2ac2b1212eefe7f748a6231184

sprunk:
There is no `weapontype` field in WeaponDefs, the previous was kosher all along though.
thx [BoS]nixtux!


abma:
i agree, this looks cleaner. thanks!

sprunk:
It's not about being cleaner! The previous didn't work.

PR #572, opened 2022-03-27
Trigger update of pieceMatrices uniform from Lua

CommanderSpice:
I did not find a way to update a UniformMatrix array from Lua.
If this is indeed not possible, redrawing a unit from Lua with `gl.Unit(unitID, true)` is broken because the piece matrices are needed in the shader.
This patch avoids pushing the big array from C++ to Lua and back again, but does not allow the updating of arbitrary UniformMatrix arrays.

Example usage to redraw a unit with the currently selected shader:
```
gl.UniformMatrix(viewLocation, "view")
gl.UniformMatrix(projLocation, "projection")
gl.UniformMatrix(modelLocation, Spring.GetUnitTransformMatrix(unitID))
gl.UniformMatrix(pieceLocation, "pieceMatrices", unitID);
gl.Unit(unitID, true)
```

### DIFF
--- a/rts/Lua/LuaShaders.cpp
+++ b/rts/Lua/LuaShaders.cpp
@@ -11,6 +11,7 @@
 #include "LuaUtils.h"
 
 #include "Game/Camera.h"
+#include "Sim/Units/UnitHandler.h"
 #include "System/Log/ILog.h"
 #include "System/StringUtil.h"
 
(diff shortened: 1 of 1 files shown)


PicassoCT:
? sprung ?
? gajop ?
? hoko ?

abma:
thank you!

PR #573, opened 2022-05-22
fix wrong BuildDistance value in AI interface

neoedmund:
location: `AI interface` / `UnitDef_getBuildDistance()`

before:
```
	test BotLab build distance=128.0
	test Lazarus build distance=96.0
	test ConstructionTurret build distance=400.0
	test Tick build distance=128.0
```

```
	test BotLab build distance=0.0
	test Lazarus build distance=96.0
	test ConstructionTurret build distance=400.0
	test Tick build distance=0.0
```

description:
128.0 is a wrong value in any point of view, set to 0 should be right.

83d3d77afe


neoedmund:
but make existed things broken, like not build things in factory, revert.

PR #574, opened 2022-06-20
Fix compile error

Elpersonn:
Compiling with the preprocessor code causes a compile error
```
/home/elperson/Downloads/spring-error/rts/Lua/LuadObjectRendering.h: In static member function ‘static int LuaObjectRendering<T>::SetForwardMaterialUniform( lua_State*)’:
/home/elperson/Downloads/spring-error/rts/Lua/LuaObjectRendering.h:168:82: ‘static int LuaObjectRenderi
ngImpl::SetForwardMaterialUniform(lua_State*)’ is private within this context
```
I am running Arch linux and my gcc version is 12.1.0


abma:
thank you!

PR #578, opened 2022-10-05
Reenable yield in springtime for short durations

dzosz:
Enables thread yielding instead of sleep for micro durations that OS probably won't schedule on time. Not sure if this change is welcome as it changes default behavior.
State before:
* buggy DEAD that enables thread yielding with influence of previous sleep times (likely just a faulty logic that was never used). yielding logic was was dead because initial avg_latency values were set to 0.

State after:
* yield thread instead of sleeping for very short durations that take no longer than 5 yield iterations (configurable). initial boundary set to 100us that gets smoothed to actual duration of yield
* At this point of time there's no code in spring that attempts to sleep for such a short duration so it is potential improvement after all


dzosz:
mistake

PR #581, opened 2022-10-20
Implement Spring.Draw to draw from Lua

CommanderSpice:
Use an unsynchronized vertex buffer to submit primitives.
OpenGL blending settings and bound textures are used.

Migration of built-in widget selector from OpenGL 3 to Spring.Draw is shown [here](/CommanderSpice/spring/commit/3e570d074a17d4c4523e0bafda2ea9aa676e8bc1#diff-7fd3ae2d1e3da477cf2c1c149f9db726f205f461156f98b67be64e47a063789aL320).

### Words of warning
- Even with this merged, there are still issues remaining in 106. Probably **a lot**. This is just one step towards a usable 106.
- Rendering rectangles in the fragment shader might become expensive. It performed better in my tests than fiddling with bevels, radii and borders on the CPU.
- This was developed on Linux with a GeForce RTX 2080 Ti using the proprietary driver. I am very interested in tests in different environments.
- Drawing during the menu phase is untested and most likely not working correctly.

### Demo widget
The [demo widget](/CommanderSpice/spring/blob/3e570d074a17d4c4523e0bafda2ea9aa676e8bc1/cont/examples/Widgets/dbg_draw_demo.lua) has commands for repeated drawing, so the impact on FPS can be measured.
Each square is drawn by just one Lua call.

draw_demo

### Wide lines
Spring.Draw.Lines draws wide lines. By modifying two lines of `cmd_mex_snap.lua`:

    gl.Texture("bitmaps/arrow.png")
    Spring.Draw.Lines(bx, by, bz, bestPos[1], bestPos[2], bestPos[3], {width=25, color={1,1,0,0.5}, animationspeed=2})

we get this:
fancy_mex_snap

This is just to show off what Spring.Draw can do, it is probably too busy to use it in a real game.


abma:
great, thank you!

PR #583, opened 2022-12-12
Vertices, fonts and fixes

CommanderSpice:
- `Spring.Draw.Vertices` submits ~~arbritrary~~ float vertex attributes to a custom shader.
- Fonts can be created as relative fonts. Sizes of relative fonts are relative to screen diagonal of 1920x1080. Relative fonts are reloaded on view port resize.
- Various fixes.


abma:
looks good, thank you!

PR #586, opened 2023-01-02
Set interface include directory for native skirmish AI targets

CommanderSpice:
The native 'compilable' skirmish AI targets have type 'MODULE_LIBRARY', Java ones have type 'UTILITY'.


abma:
CircuitAI was removed: 

https://github.com/spring/spring/commit/dd4d9de24ca47b94fea2b2d63700d39f3a4de606

do you think its useful to keep it atm as its unmaintained for spring and it heavily increases compile time + download size?

the others AIs don't have the problem with the missing include it seems.

IMHO the target_include_directories should be changed/added here:

https://github.com/spring/CircuitAI/blob/master/CMakeLists.txt#L48

CommanderSpice:
My bad, I missed that. I'll create a PR there.

I am fine with dropping it for now.

PR #587, opened 2023-01-05
Fix ground circles; Drop OpenGL 3 matrix mode check

CommanderSpice:
I thought about replacing the matrix mode check with a check of the application matrix state `GLMatrixStateTracker`, but I think that `GLMatrixStateTracker` should be removed completely in the future because we are not trying to  provide the OpenGL 3 API to widgets.


abma:
thank you!

PR #590, opened 2023-03-31
add some OpenBSD codepaths, mostly following FreeBSD

rfht:
This is some of the low-hanging fruit to help work towards running spring on OpenBSD. Mostly follows the FreeBSD codepaths. With these (and a handful more selected changes), I'm able to launch Spring on OpenBSD. Some OpenBSD specifics are the use of `getthrid()` for thread id, and that we don't have a separate librt (similar to Apple, apparently).


abma:
thank you!

PR #592, opened 2023-06-10
Added Brackets to for loop 

SageTheWizard:
For loop threw compiler warning due to misleading indentation  on for loop
The 2nd If statement would not be executed within loop, added brackets so that the final if statement would be executed


abma:
please split code style changes and code changes into individual commits: else its impossible to review the code.

also zip.c is an "external" file: these changes should be done upstream and then merged back.

PR #596, opened 2023-12-27
Add missing cstdint headers

alexs-sh:
Hello,

during building with gcc13, I faced several issues caused by missing cstdint header. Here are the changes that allowed me to finish the build. Feel free to cancel if gcc13 support is not planned or ask for update/rework changes.

Thank you


abma:
thank you!

